PR #49, opened 2023-08-29
One-off query support

kazimuth:
## Description of Changes
Adds an async one-off-query method to SpacetimeDBClient.

## API
Adds:
```cs
public async Task<T[]> SpacetimeDBClient.OneOffQuery<T>(string query) where T : IDatabaseTable;
```

## Requires SpacetimeDB PRs
https://github.com/clockworklabs/SpacetimeDB/pull/198


src/SpacetimeDBClient.cs
```diff
@@ -7,6 +7,7 @@
 using System.Reflection;
 using System.Text;
 using System.Threading;
+using System.Threading.Channels;
```
jdetter, line 10: I had to add a nuget package in order to get this working, did you not have to do this?

jdetter (approved):
Take a look at my comment before merging, but I was able to get this working.

src/SpacetimeDBClient.cs
kazimuth: Not with `dotnet build`, but it's possible that pulled in some floating dependency from my system?

jdetter:
Merging this into `staging`, this will go into `v0.7.0-beta`

PR #51, opened 2023-09-11
Multiple fixes for SomeWrapperConverter

jdetter:
## Description of Changes

 - Fixes the namespace for the `SomeWrapper` class so that it matches the expected output by the CLI.

## API

## Requires SpacetimeDB PRs


bfops:
( @jdetter is this still relevant? )

PR #65, opened 2023-12-21
Performance Fix - Locking in Update

jdetter:
## Description of Changes

- Fixes a performance issue which was introduced by acquiring a lock in `Update()`

## API

## Requires SpacetimeDB PRs


jdetter:
This seems to have been caused by an issue in bitcraft, this code doesn't actually fix anything

PR #70, opened 2024-02-06
More than one network message is processed per update call

jdetter:
## Description of Changes

- Fixes an issue where an update call can only process a single network message.

## API

## Requires SpacetimeDB PRs
None


jdetter:
This has already been tested via BitCraft

PR #72, opened 2024-02-20
Remove the `rowPk` from the client API

gefjon:
## Description of Changes

Re: https://github.com/clockworklabs/SpacetimeDB/pull/840

This commit updates the C# SDK to no longer use the `row_pk` field of the Protobuf client API,
as that field has been removed. (Will have been removed, as of merging.)

Where a table cache was previously keyed on `byte[] rowPk`, it is now keyed on `byte[] rowBytes`,
where `rowBytes` is the BSATN-encoded bytes of the row. This means we effectively store two copies of each row in the client cache: the BSATN serialized format, and the decoded domain type.

An alternate implementation would be to make the table caches be sets of domain types, discarding the BSATN bytes.
We find this undesirable for several reasons:
- Hashing and equality-comparing `byte[]` is almost certainly more efficient than doing the same for domain types.
- Even if hashing and equality-comparing domain types were efficient, we would still have to update codegen to emit hashing and equality methods for all types in the module_bindings. This implementation requires no changes to the module_bindings.
- We already have the BSATN bytes sitting around, as they're necessarily part of the message we recieve from the server. This change does no additional serialization or deserialization.

In essence, we're trading memory for time and simplicity. Keeping the BSATN bytes live approximately doubles the table cache's memory usage, but simplifies the implementation greatly,
and (we suspect) speeds up table cache insertions, deletions and lookups.

Also, this PR changes the quickstart-chat example to depend on the SDK by `ProjectReference`, rather than `PackageReference`. This allows `dotnet build` in `examples/quickstart/client` to build against local changes to the SDK, rather than against a pinned NuGet version. (The pinned NuGet version was woefully outdated.)

## API

 - [ ] This is a breaking change to the SDK's user-facing API
 - [x] This is a breaking change to the internal API which the SDK uses to communicate with SpacetimeDB 

## Requires SpacetimeDB PRs

- https://github.com/clockworklabs/SpacetimeDB/pull/840


jdetter (approved):
Tested with phoebe during meet, works! 👍

PR #78, opened 2024-04-22
Merge ByteArrayComparers

RReverser:
## Description of Changes

There were couple of mismatched implementations of byte array comparison / hashing. This PR simply keeps the most efficient one as a utility and forwards all the usages to it.

## API

## Requires SpacetimeDB PRs

### DIFF
--- a/src/Utils.cs
+++ b/src/Utils.cs
@@ -1,44 +1,51 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+using System.Runtime.CompilerServices;
 
 namespace SpacetimeDB
 {
-    public static class Utils
+    internal readonly struct ByteArrayComparer : IEqualityComparer<byte[]>
     {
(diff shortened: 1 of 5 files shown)


kazimuth (approved):
LGTM.

We might wanna just run a formatter on all this after we get the rest of your PRs in to deal with any excess whitespace etc.

src/Utils.cs
```diff
@@ -1,44 +1,51 @@
...
-            int sum = 0;
-            for (int i = 0; i < obj.Length; i++)
-                sum += obj[i];
-            return sum;
+            int hash = 17;
+            foreach (byte b in obj)
+            {
+                hash = hash * 31 + b;
+            }
+            return hash;
```
kazimuth, line 48: We could think about pulling in a fast hash as a dependency if needed, but I think this should work fine.

RReverser:
@jdetter Was a bit concerned because this is touching performance-sensitive code and wanted to test it with Bitcraft, so leaving open until his approval.

PR #79, opened 2024-04-22
Make logger a separate SDK setting from client instance

RReverser:
## Description of Changes

Extracts logger into a separate global setting from the client itself.

This allows to:

1. Have a preset logger that is already correctly matching the environment (console vs Unity).
2. Removes the need for explicit `SpacetimeDBClient.CreateInstance(...)` which is particularly awkward to use and easy to forget with singleton as it doesn't return a result as the name (which looks like a factory method) could suggest. Instead, `SpacetimeDBClient.instance` is now automatically available on first use.
3. Slightly simplifies dependencies between classes, e.g. `ClientCache` no longer needs a circular dependency on `SpacetimeDBClient`, making future maintenance and changes a bit easier.

## API

## Requires SpacetimeDB PRs


jdetter (approved):
LGTM as long as it works in Unity 👍

PR #81, opened 2024-04-23
Extract common logic from `OnMessageProcessComplete`

RReverser:
## Description of Changes

Minor refactoring that extracts common part of subscription and transaction update handling into a separate function, so that the main logic in `OnMessageProcessComplete` is a bit more straightforward and handles all event types with a sinlge `switch` instead of 3 separate `switch`es.

This is a small no-op step that makes further refactoring and optimisation of said common logic a bit easier.

## API

## Requires SpacetimeDB PRs

### DIFF
--- a/src/SpacetimeDBClient.cs
+++ b/src/SpacetimeDBClient.cs
@@ -286,7 +286,7 @@ namespace SpacetimeDB
                 {
                     if (!subscriptionInserts.TryGetValue(tableName, out var hashSet))
                     {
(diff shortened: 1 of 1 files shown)


kazimuth (approved):
Again, this looks good. It seems like a reasonably straightforward refactor.

RReverser:
@jdetter Wanted to test this one with Bitcraft due to relatively large-looking diff, so leaving open until his approval.

PR #82, opened 2024-04-24
Fix nullability warnings in WebSocket class

RReverser:
## Description of Changes

There is a bunch of nullability issues in the project; this PR fixes just the ones in WebSocket because a lot of other nullability issues are in code that's going to be removed in BSATN implementation.

Also contains some drive-by fixups / removing unused fields / syntax simplifications in WebSocket class as well.

## API

## Requires SpacetimeDB PRs

### DIFF
--- a/src/WebSocket.cs
+++ b/src/WebSocket.cs
@@ -1,6 +1,6 @@
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
+using System.Linq;
 using System.Net.WebSockets;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


kazimuth (approved):
All of this is supported by Unity, right? I've run into issues trying to get rid of nullability warnings doing stuff like this before.

Also, if needed you can use an EditorConfig file to disable the un-fixable nullability warnings: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/suppress-warnings 

I had a branch where I did this but it seems to have been lost to the ether.

RReverser:
> All of this is supported by Unity, right? I've run into issues trying to get rid of nullability warnings doing stuff like this before.

Yup, definitely supported.

> Also, if needed you can use an EditorConfig file to disable the un-fixable nullability warnings

You can just use `#nullability disable` at the top of the file instead - which is something I did in my branch where all files except one have nullability issues fixed. That's a bit easier than separate config.

RReverser:
@jdetter I'm merging this one with just @kazimuth's review as I think it's mostly pretty straightforward / trivial.

PR #84, opened 2024-05-01
Minor refactoring of AuthToken

RReverser:
## Description of Changes

Initially just fixed the nullability issues / warnings, but in the process did a few more minor drive-by refactorings.

## API

## Requires SpacetimeDB PRs


jdetter (approved):
LGTM

PR #85, opened 2024-05-09
NFC: remove unused primaryKeyValue field

RReverser:
## Description of Changes

This field is only used in self-assignment, so can be safely removed.

## API

## Requires SpacetimeDB PRs


jdetter (approved):
LGTM

PR #86, opened 2024-05-10
NFC: inline insert/delete update handling

RReverser:
## Description of Changes

Instead of a separate TableOp enum, and 4 fields to handle insert/delete metadata in DbOp, use a nullable struct that contains non-nullable object + bytes pair internally:

- if `insert` is present (non-nullable) and `delete` is nullable, that naturally indicates insert operation
- if `insert` is nullable and `delete` is not, that's a delete
- if both are non-nullable, it's an update
- if both are null, it's an internal "no change" state

This simplifies and shortens update handling as well as reduces risk of state getting out of sync - e.g. TableOp saying that the operation is insert should exist but field containing inserted value being null. Now nullable struct itself communicates whether there is an insert or not, without a separate enum.

## API

## Requires SpacetimeDB PRs

### DIFF
--- a/src/SpacetimeDBClient.cs
+++ b/src/SpacetimeDBClient.cs
@@ -20,14 +20,6 @@ namespace SpacetimeDB
 {
     public class SpacetimeDBClient
     {
-        public enum TableOp
-        {
-            Insert,
-            Delete,
-            Update,
-            NoChange,
-        }
-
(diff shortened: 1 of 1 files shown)


jdetter (approved):
LGTM - seems like a generally good simplification 👍

PR #83, opened 2024-05-01
Remove RowUpdate event

RReverser:
## Description of Changes
*Describe what has been changed, any new features or bug fixes*

We talked about this for a while, so decided to do this in a separate PR.

## API

 - [x] This is an API breaking change to the SDK

*If the API is breaking, please state below what will break*

Users will need to subscribe to OnInsert / OnUpdate / OnDelete separately.

## Requires SpacetimeDB PRs
*List any PRs here that are required for this SDK change to work*

https://github.com/clockworklabs/SpacetimeDB/pull/1190

PR #87, opened 2024-05-11
Fix conflict resolution shenanigans

RReverser:
## Description of Changes

Looks like something went terribly wrong with conflict resolution during rebase of https://github.com/clockworklabs/spacetimedb-csharp-sdk/pull/83 (in particular, a missing brace), and we don't have CI to catch it so it got merged as-is.

## API

## Requires SpacetimeDB PRs


jdetter (approved):
LGTM

PR #88, opened 2024-05-11
Add basic CI

RReverser:
## Description of Changes

As found out the hard way in https://github.com/clockworklabs/spacetimedb-csharp-sdk/pull/87, currently there are no CI checks to validate that code is even syntactically valid before merging.

This adds a basic build step that verifies that client example (and, by extension, the library itself as it relies on it) builds successfully.

## API

## Requires SpacetimeDB PRs


jdetter (approved):
Love to see it, thanks Ingvar 👍

PR #89, opened 2024-05-12
Switch to TaskCompletionSource

RReverser:
## Description of Changes

There is only one place where we use Channels, and it's to create and await a channel with one element - which is functionally the same as a more precise and low-weight TaskCompletionSource.

Switching also makes the SDK compatible with the widely supported .NET Standard 2.1 subset, which is supported natively in Unity and allows to remove a custom System.Threading.Channels package from Unity SDK dependencies.

## API

## Requires SpacetimeDB PRs


jdetter (approved):
LGTM - let's remember to delete the custom threading package in the Unity SDK.

PR #93, opened 2024-05-22
Simplify update handling as per consistent filtering proposal

RReverser:
## Description of Changes

As a result of the consistent filtering rules proposal (clockworklabs/SpacetimeDB#1256), all equatable types are now natively equatable in C# as well. 

This allows us to compare objects directly, without holding and comparing AlgebraicValue in the entries map as well, which has a bit of a domino effect and allows to optimise, simplify or even remove some parts of the SDK.

## API

New filtering rules limit types on which filtering can be done, as well as change the return type of `FilterBy` functions to always be iterable.

## Requires SpacetimeDB PRs

https://github.com/clockworklabs/SpacetimeDB/pull/1277

### DIFF
--- a/src/SpacetimeDBClient.cs
+++ b/src/SpacetimeDBClient.cs
@@ -290,6 +290,19 @@ namespace SpacetimeDB
                             foreach (var row in update.TableRowOperations)
                             {
                                 var rowBytes = row.Row.ToByteArray();
+
+                                if (row.Op != TableRowOperation.Types.OperationType.Insert)
+                                {
+                                    Logger.LogWarning("Non-insert during a subscription update!");
+                                    continue;
+                                }
+
+                                if (!hashSet.Add(rowBytes))
+                                {
+                                    // Ignore duplicate inserts in the same subscription update.
+                                    continue;
+                                }
+
                                 stream.Position = 0;
                                 stream.Write(rowBytes, 0, rowBytes.Length);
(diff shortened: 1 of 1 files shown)


examples/quickstart/client/module_bindings/Message.cs
```diff
@@ -43,9 +43,7 @@ public static SpacetimeDB.SATS.AlgebraicType GetAlgebraicType()
 
 		public static explicit operator Message(SpacetimeDB.SATS.AlgebraicValue value)
 		{
-			if (value == null) {
-				return null;
-			}
```
jdetter, line 48: I know its somewhat controvertial but our style guide typically requires the braces here

jdetter (approved):
Small nit but overall this is a good refactor. I have not tested it myself but will be doing an overall test of the C#/Unity SDK later this week so this should get tested as part of that.

examples/quickstart/client/module_bindings/Message.cs
RReverser: I think it got changed in one of the previous PRs, it's just that module bindings weren't regenerated. If we want to revert, probably worth doing outside of this PR.

RReverser: Ah, I'm actually going to remove that function altogether in the next PR, so hopefully that solves it too.

PR #95, opened 2024-05-23
Bump version to 0.9.2

bfops:
## Description of Changes
Just bumps the version number. Since 0.9.1 has been released, we are now on 0.9.2+!

## API

## Requires SpacetimeDB PRs

### DIFF
--- a/SpacetimeDB.ClientSDK.csproj
+++ b/SpacetimeDB.ClientSDK.csproj
@@ -18,7 +18,7 @@
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <RepositoryUrl>https://github.com/clockworklabs/spacetimedb-csharp-sdk</RepositoryUrl>
     <AssemblyVersion>0.8.0</AssemblyVersion>
-    <Version>0.9.0</Version>
+    <Version>0.9.2</Version>
   </PropertyGroup>
 
   <ItemGroup>


RReverser:
Did we land any new changes for that one? If not, I'm pretty sure that back when we talked about it, Tyler said we don't need to keep patch versions the same across SDKs because each SDK might have its own changes / development not related to SpacetimeDB itself.

So we set 0.x to match across SDKs but last patch component is for each SDKs own iteration.

RReverser:
As discussed in chat, let's either bump one up (to 0.9.1) or just not bump at all since master already has a tag so no new changes here.

kurtismullins:
> As discussed in chat, let's either bump one up (to 0.9.1) or just not bump at all since master already has a tag so no new changes here.

I think that moving forward, this is a good idea. Since @bfops already had the versions set to v0.9.2-beta, I decided to leave it there for this release. I will not merge this Pull Request so that `master` is unaffected. Just note that the latest version pushed to nuget is v0.9.2.

As mentioned in Discord -- we did skip v0.9.1. Maybe someone can just push a release there so its not missing? It can be the same as v0.9.0. I don't think its super important but does look weird, historically speaking.

RReverser:
> I will not merge this Pull Request so that `master` is unaffected. Just note that the latest version pushed to nuget is v0.9.2.

Well if it's published to nuget, then let's merge it. Github repo should reflect nuget state, or the next person trying to bump version is gonna be even more confused.

kurtismullins (approved):

RReverser:
In fact, we should have merged it before I merged https://github.com/clockworklabs/spacetimedb-csharp-sdk/commit/025a84cded5f38a050262c8afaacfafd8169f233 as the history looks weird now (that commit is not part of 0.9.2, even though it precedes it). Oh well...

PR #96, opened 2024-05-23
Release/candidate/v0.9.2

bfops:
## Description of Changes

## API

## Requires SpacetimeDB PRs


kurtismullins (approved):
This has been pushed to nuget. I am going to approve + merge to get the version number back into the mainline branch. (in this case, its `release/base` but we also need to get it into `master` for future commits).

PR #100, opened 2024-06-18
Switch to variadic params for Subscribe queries

RReverser:
## Description of Changes

Accept variadic params or array of queries instead of a `List<string>`.

This is a breaking change, but it should be a slight improvement for the majority of cases with statically defined queries.

## API

## Requires SpacetimeDB PRs


RReverser:
No longer necessary with the new subscription API.

PR #110, opened 2024-07-23
CI - `pull_request` events don't also trigger on `push`

bfops:
## Description of Changes
Currently, a push to a PR triggers our `dotnet` workflows twice: once for the `push` event, and once for the `pull_request` event.

This PR sets the `pull_request` events to only be `opened` and `reopened`, so that a push to the branch only triggers the `push` event and not the PR `synchronize` event as well.

## Testing
- [x] CI has only triggered once on this PR


.github/workflows/dotnet.yml
```diff
@@ -3,7 +3,11 @@
 
 name: .NET build
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - master
```
bfops, line 10: this matches the behavior in SpacetimeDB (https://github.com/clockworklabs/SpacetimeDB/blob/master/.github/workflows/ci.yml)

RReverser: I think SpacetimeDB has somewhat different constraints because it's so large / expensive to build, whereas this repo takes less than a minute.

I love being able to run CI on branches even before I make a PR. Could we preserve that?

bfops: updated to run on `push` events but _not_ `ssynchronize` events

RReverser (approved):
I don't fully understand the change in behaviour after this PR, but sounds good.

.github/workflows/dotnet.yml
```diff
@@ -3,7 +3,10 @@
 
 name: .NET build
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+    types: [opened, reopened]
```
bfops, line 9: The default (i.e. if `types` isn't specified) is `[opened, reopened, synchronized]`.

The `synchronized` event happens whenever commits are pushed to a PR, which was redundant with us also triggering on all `push`es.

See also https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

bfops:
> I don't fully understand the change in behaviour after this PR, but sounds good.

@RReverser Added a hopefully-clarifying comment here: https://github.com/clockworklabs/spacetimedb-csharp-sdk/pull/110/files#r1693270679

.github/workflows/dotnet.yml
RReverser: > The `synchronized` event happens whenever commits are pushed to a PR, which was redundant with us also triggering on all `push`es.

IIRC they do that intentionally because the commit on the branch itself might be different from a merge commit when PR is out of date (not rebased), so the test result might be different as well? Admittedly, it's an edge case, but sometimes it might potentially catch issues early.

bfops: Good point.. but it seems like this repo requires PRs to be up to date with master before merging, so I think it might not apply here?

bfops: actually.. based on https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request, it looks like `synchronize` is probably only triggered when the head branch is pushed.

I think there might be a different option/event type for testing the merge commits? IIRC they're pushed to some other ref that gets CI tested.

RReverser: Oh huh. I probably confused the event types. But yeah, either way the base branch requirement takes care of that.

PR #111, opened 2024-07-25
Minor BSATN simplifications

RReverser:
## Description of Changes

Minor simplifications using BSATN helpers for Protobufectomy code.

Also removed unnecessary usings and a pattern.

## API

## Requires SpacetimeDB PRs

### DIFF
--- a/src/WebSocket.cs
+++ b/src/WebSocket.cs
@@ -1,8 +1,8 @@
+using SpacetimeDB.BSATN;
 using SpacetimeDB.ClientApi;
 
 using System;
 using System.Collections.Concurrent;
-using System.IO;
 using System.Linq;
 using System.Net.WebSockets;
 using System.Text;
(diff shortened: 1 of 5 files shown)


src/WebSocket.cs
```diff
@@ -185,10 +185,9 @@ private async Task ProcessSendQueue()
                         }
                     }
 
-                    var o = new MemoryStream();
-                    var w = new BinaryWriter(o);
-                    new ClientMessage.BSATN().Write(w, message);
-                    await Ws!.SendAsync(o.ToArray(), WebSocketMessageType.Binary, true, CancellationToken.None);
+                    var messageBSATN = new ClientMessage.BSATN();
+                    var encodedMessage = IStructuralReadWrite.ToBytes(messageBSATN, message);
+                    await Ws!.SendAsync(encodedMessage, WebSocketMessageType.Binary, true, CancellationToken.None);
```
bfops, line 190: do the current tests exercise this codepath, or should it be tested via BitCraft?

RReverser: Unfortunately, the tests don't use networking, they're pure unit tests. I don't expect this to be a functional change as that helper does the same thing, but I guess worth testing.

bfops (approved):
Seems fine to me!

This does look to be identical based on the definitions of `ToBytes` in `crates/bindings-csharp-BSATN/Runtime/BSATN/Runtime.cs`

PR #116, opened 2024-07-31
Don't use BuiltinType.I/U128, use AlgebraicType.I/U128

Centril:
## Description of Changes

Required to make "SDK Tests" pass in https://github.com/clockworklabs/SpacetimeDB/pull/1559.

## API

Not breaking.

## Requires SpacetimeDB PRs

- https://github.com/clockworklabs/SpacetimeDB/pull/1559


bfops (approved):
Seems hard to argue with and it passes CI!

PR #118, opened 2024-08-06
Restore `dotnet pack` functionality

bfops:
## Description of Changes
Single-line change so that `dotnet pack` stops complaining that nothing was generated.

Per @RReverser this brings this package more in line with our other C# packages.

## API
Nah nothing breaking.

## Requires SpacetimeDB PRs
Nope

### DIFF
--- a/SpacetimeDB.ClientSDK.csproj
+++ b/SpacetimeDB.ClientSDK.csproj
@@ -9,7 +9,6 @@
     <Title>SpacetimeDB SDK</Title>
     <Authors>jdetter</Authors>
     <Company>Clockwork Labs</Company>
-    <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
     <Product>SpacetimeDB</Product>
     <Description>The SpacetimeDB SDK is a software development kit (SDK) designed to simplify the interaction with SpacetimeDB server modules.</Description>
     <Copyright>2024</Copyright>
@@ -31,7 +30,6 @@
   <ItemGroup>
     <None Update="logo.png" Pack="true" PackagePath="" />
     <None Update="README.md" Pack="true" PackagePath="" />
-    <None Include="LICENSE.txt" Pack="true" PackagePath="LICENSE.txt" />
   </ItemGroup>
 
 </Project>
--- a/SpacetimeDB.ClientSDK.csproj
+++ b/SpacetimeDB.ClientSDK.csproj
@@ -5,11 +5,11 @@
     <LangVersion>9</LangVersion>
     <ImplicitUsings>disable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     <PackageId>SpacetimeDB.ClientSDK</PackageId>
     <Title>SpacetimeDB SDK</Title>
     <Authors>jdetter</Authors>
     <Company>Clockwork Labs</Company>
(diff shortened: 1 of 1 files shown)


SpacetimeDB.ClientSDK.csproj
```diff
@@ -5,11 +5,11 @@
     <LangVersion>9</LangVersion>
     <ImplicitUsings>disable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     <PackageId>SpacetimeDB.ClientSDK</PackageId>
     <Title>SpacetimeDB SDK</Title>
     <Authors>jdetter</Authors>
     <Company>Clockwork Labs</Company>
+    <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
```
RReverser: Maybe better to use `PackageLicenseExpression` so that it can nicely show up on NuGet pages?

bfops: it sounds like there isn't an expression we could use.. and also it's currently intentional that we publish without a license, so I'm removing these

RReverser: Hm LICENSE.txt looks like a standard `Apache-2.0` SPDX?

SpacetimeDB.ClientSDK.csproj
```diff
@@ -5,7 +5,6 @@
     <LangVersion>9</LangVersion>
     <ImplicitUsings>disable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
```
RReverser, line 8: Is this removed just because we don't need it for tests and such?

bfops: it prevented `dotnet pack` from working properly and this fixed it, I have no context on what it does :shrug: 
https://steveellwoodnlc.medium.com/error-nu5026-the-file-to-be-packed-was-not-found-on-disk-18bfbc6be4a

SpacetimeDB.ClientSDK.csproj
bfops: apparently we currently intentionally publish with no license

bfops: (I assume since I've reverted these changes then `LICENSE.txt` won't be bundled in?)

SpacetimeDB.ClientSDK.csproj
RReverser: Weird. This line basically integrates packing right into `dotnet build` so that you don't have to run `dotnet pack` separately... but I had no idea it might conflict with the latter.

Removing sgtm.

RReverser (approved):

SpacetimeDB.ClientSDK.csproj
bfops: oh! then I can just update the release docs to say run `dotnet build --configuration Release` instead. looks like it works for me!

RReverser: Noo, I was actually fine with removing that line for consistency with other repos 😅

Also I suspect it's somewhat inefficient now that we have tests - because it will probably build the NuGet package even for `dotnet test`, which is kinda wasteful.

bfops: restored!

PR #131, opened 2024-09-12
c# client SDK (DbConnection + SDK Callbacks)

lcodes:
Implementation of the db connection proposal for C#


RReverser:
You might want to rebase & retarget this against the `staging` branch that has some of the latest changes not yet on `master`.

src/Stubs.cs
```diff
@@ -31,4 +46,20 @@ public ReducerEventBase(TransactionUpdate update)
 
         public abstract bool InvokeHandler();
     }
+
+    public abstract class RemoteBase<DbConnection> {
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor.
```
RReverser: Why not just turn `Init` into an actual constructor?

lcodes: Ah hmm I'll recheck that one, I think I initially put that there because of the new() constraint

src/Stubs.cs
```diff
@@ -31,4 +46,18 @@ public ReducerEventBase(TransactionUpdate update)
 
         public abstract bool InvokeHandler();
     }
+
+    public abstract class RemoteBase<DbConnection> {
+        protected DbConnection conn;
```
RReverser: Nit: can this one be `readonly`?

src/SpacetimeDBClient.cs
```diff
@@ -126,7 +160,7 @@ struct ProcessedMessage
             public ServerMessage message;
             public List<DbOp> dbOps;
             public DateTime timestamp;
-            public ReducerEvent? reducerEvent;
+            public EventContext reducerEvent;
```
RReverser: Does this one really not need to be nullable anymore? I think there was some case in which a reducer we returned was `null` (if it's unknown type maybe?).

examples~/quickstart/client/spacetimedb-quickstart-client-cs.csproj
```diff
@@ -11,6 +11,8 @@
 
   <ItemGroup>
     <ProjectReference Include="../../../SpacetimeDB.ClientSDK.csproj" />
+    <ProjectReference Include="..\..\..\..\stdb-module-cs\crates\bindings-csharp\BSATN.Codegen\BSATN.Codegen.csproj"  OutputItemType="Analyzer" />
```
RReverser: Those paths will need to be fixed (removed?) before merging.

examples~/quickstart/client/module_bindings/_Globals/SpacetimeDBClient.cs
```diff
@@ -10,59 +10,92 @@
...
-	public class SpacetimeDBClient : SpacetimeDBClientBase<ReducerEvent>
+	public class DbConnection : DbConnectionBase<DbConnection, EventContext>
 	{
-		protected SpacetimeDBClient()
+		public readonly RemoteTables RemoteTables = new();
+		public readonly RemoteReducers RemoteReducers = new();
+
+		public DbConnection()
 		{
+			RemoteReducers.Init(this);
```
RReverser: Should this one be part of the constructor invocation now too? (since you changed  RemoteBase signature)

lcodes: Ah yeah forgot to rerun codegen

src/SpacetimeDBClient.cs
lcodes: Good point, Ill check

src/SpacetimeDBClient.cs
```diff
@@ -14,9 +14,64 @@
...
+            return this;
+        }
+
+        public DbConnectionBuilder<DbConnection, EventContext> OnConnect(Action<Identity, string> cb) {
+            conn.onConnect += cb;
+            return this;
+        }
+
+        public DbConnectionBuilder<DbConnection, EventContext> OnConnectError(Action<WebSocketError?, string> cb) {
+            conn.webSocket.OnConnectError += (a, b) => cb.Invoke(a, b);
```
RReverser, line 69: Is this not the same as

```suggestion
            conn.webSocket.OnConnectError += cb;
```

RReverser:
> You might want to rebase & retarget this against the `staging` branch that has some of the latest changes not yet on `master`.

@lcodes Also in case you missed this one above. This should exclude some unrelated changes from your PR (also we just shouldn't merge into `master`).

src/SpacetimeDBClient.cs
lcodes: No, it cannot convert the Action<> into the Handler delegate even if the arguments are the same.

RReverser: I guess you could accept the handler type here?

RReverser: Not important anyway.

src/Stubs.cs
```diff
@@ -1,23 +1,38 @@
...
+        public ReducerView Reducers;
+
+        public DbContext(DbView db, ReducerView reducers) : base(db) => Reducers = reducers;
+    }
+
+    public interface IEventContext {
+        bool InvokeHandler();
+    }
+
+    public abstract class EventContextBase<RemoteTables, RemoteReducers> : DbContext<RemoteTables, RemoteReducers>, IEventContext
```
RReverser: Just in case - this doesn't seem to match the structure described in the "SDK callbacks" section of the proposal - are those changes left for a separate PR?

SpacetimeDB.ClientSDK.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
RReverser, line 1: I wonder what's constantly going on with those. Was there a different newline or something?

src/Stubs.cs
```diff
@@ -1,23 +1,38 @@
...
+        where RemoteTables : class, new()
+        where RemoteReducers : class, new()
```
RReverser: This bound doesn't work anymore. CI would probably catch it, but it doesn't get that far.

This class doesn't need the `new()` constraints though, or, for that matter, any constraints for those generics, since it accepts them from outside.

examples~/quickstart/client/module_bindings/_Globals/SpacetimeDBClient.cs
```diff
@@ -10,59 +10,94 @@
...
 	{
-		protected SpacetimeDBClient()
+		public readonly RemoteTables RemoteTables = new();
+		public readonly RemoteReducers RemoteReducers;
+
+		public DbConnection()
 		{
+			RemoteReducers = new(this);
+
 			clientDB.AddTable<Message>();
```
RReverser: Should the client cache be migrated over to use RemoteTables now?

src/Stubs.cs
```diff
@@ -1,23 +1,38 @@
...
+
 namespace SpacetimeDB
 {
     public interface IReducerArgsBase : BSATN.IStructuralReadWrite
     {
         string ReducerName { get; }
     }
 
-    public abstract class ReducerEventBase
+    public abstract class DbContext<DbView, ReducerView> : DbContext<DbView>
```
RReverser, line 12: This class definitely can be merged into `EventContextBase`.

src/Stubs.cs
RReverser: (fixed this one myself)

examples~/quickstart/client/module_bindings/_Globals/SpacetimeDBClient.cs
```diff
@@ -10,59 +10,94 @@
...
+			);
+			return true;
+		}
+	}
+
+	public partial class EventContext : EventContextBase<RemoteTables, RemoteReducers>
+	{
+		public IReducerArgs? Args { get; }
+
+		public string ReducerName => Args?.ReducerName ?? "<none>";
```
gefjon: This is not as designed - `EventContext` should have a single field `event: Event<Reducer>` (where `Reducer` is an enum / tagged union defined in the codegen).

RReverser: Yes, those parts of the proposal are not covered in this PR - this is struct as it looked before. Jeremie is working on other parts separately.

gefjon: Can y'all update the PR title and description to reflect that, then? Puru was confused by this and wasted time in his implementation of the TypeScript version because he was basing it off this PR.

RReverser: Yeah I was a bit confused by it too, see my comment above. https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/131#discussion_r1771382419

@lcodes can you update the PR as suggested to clarify it's only implementing the DbConnection bits?

SpacetimeDB.ClientSDK.csproj
lcodes: maybe a byte order mark?

examples~/quickstart/client/module_bindings/_Globals/SpacetimeDBClient.cs
lcodes: Ah that's a good point.

src/Event.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+
+namespace SpacetimeDB
+{
+    public interface IEventContext { }
+
+    public interface IReducerArgs : BSATN.IStructuralReadWrite
```
RReverser, line 7: It used to have separate `IReducerArgs` per generated module, with base `IReducerArgsBase` in the library, to make sure that user doesn't accidentally mix up reducer types from different modules.

Is this no longer a risk? Are there any public APIs that rely on `IReducerArgs`?

src/SpacetimeDBClient.cs
```diff
@@ -36,26 +98,13 @@ struct DbOp
...
-        /// <summary>
-        /// Called when a connection is established to a spacetimedb instance.
-        /// </summary>
-        public event Action? onConnect;
-
-        /// <summary>
-        /// Called when a connection attempt fails.
-        /// </summary>
-        public event Action<WebSocketError?, string>? onConnectError;
+        public event Action<Identity, string>? onConnect;
```
RReverser: IIUC, these events no longer need to be public, as we configure them via the DbConnectionBuilder instead?

examples~/quickstart/client/module_bindings/_Globals/SpacetimeDBClient.cs
```diff
@@ -5,64 +5,158 @@
...
+            }
+        }
+
+        public class UserHandle : RemoteTableHandle<EventContext, User> {
+			public override object? GetPrimaryKey(IDatabaseRow row) => ((User)row).Identity;
+
+            private Dictionary<SpacetimeDB.Identity, User> Identity_Index = new(16);
+
+            public override void InternalInvokeValueInserted(IDatabaseRow row) {
+				var value = (User)row;
```
RReverser: Nit: formatting.

src/Event.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
```
RReverser, line 1: Did you add this project to test in Unity yet? It should generate `.meta` files that will need to be added to the repo.

src/Table.cs
```diff
@@ -0,0 +1,72 @@
...
+        void InvokeDelete(IEventContext context, IDatabaseRow row);
+        void InvokeBeforeDelete(IEventContext context, IDatabaseRow row);
+        void InvokeUpdate(IEventContext context, IDatabaseRow oldRow, IDatabaseRow newRow);
+    }
+
+    public abstract class RemoteTableHandle<EventContext, Row> : IRemoteTableHandle
+        where EventContext : class, IEventContext
+        where Row : IDatabaseRow, new()
+    {
+        public void SetCache(ClientCache.ITableCache cache) => Cache = (ClientCache.TableCache<Row>)cache;
```
RReverser, line 39: We can leave this as a future work, but IIUC there is no reason for TableCache and RemoteTableHandle to be separate classes?

Previously all tables were static, which is why we needed a separate class for TableCache, but now that we have handles, we can just put cached items right into them.

src/Event.cs
lcodes: It's only used with InternalCallReducer now, and we generate those calls in the table handles.

lcodes: I removed it after its main functionality moved to the generated DbConnection.Dispatch to fit the RemoteReducers; the interface was left empty.

src/SpacetimeDBClient.cs
lcodes: Oh this one is tricky; it used to be internal and only used from the builder; but the tests need a way to instantiate a DbConnection without the builder (because its Build() member creates the WS connection)

src/Table.cs
lcodes: Ah that's a good point, yeah we can leave merging them for later.

src/Event.cs
lcodes: Ah hmm I can get it to generate these files

src/SpacetimeDBClient.cs
lcodes: Ok turns out [InternalsVisibleTo] works

RReverser:
@lcodes Can unhandledReducerError, onEvent and so on be made internal as well now? Ideally hide all the APIs that are now covered elsewhere.

src/SpacetimeDBClient.cs
```diff
@@ -523,19 +572,16 @@ private void OnMessageProcessCompleteUpdate(ReducerEvent? dbEvent, List<DbOp> db
                     {
                         case { insert: { value: var newValue }, delete: { value: var oldValue } }:
                             {
-                                // If we matched an update, these values must have primary keys.
-                                var newValue_ = (IDatabaseTableWithPrimaryKey)newValue;
-                                var oldValue_ = (IDatabaseTableWithPrimaryKey)oldValue;
-                                oldValue_.OnUpdateEvent(newValue_, dbEvent);
+                                dbOp.table.Handle.InvokeUpdate(eventContext, oldValue, newValue);
```
RReverser, line 575: It's not a big deal, but FWIW a separate interface and the cast were there to ensure that we don't accidentally emit update events on tables that don't have a primary key, since it's not valid to do so.

In principle, this shouldn't happen if we reached here, but 🤷‍♂️

RReverser (approved):

examples~/quickstart/client/Program.cs
```diff
@@ -88,44 +96,59 @@ void PrintMessage(Message message)
...
+        var e = reducer.ReducerEvent;
+        if (e.CallerIdentity == local_identity && e.Status is Status.Failed(var error))
+        {
+            Console.Write($"Failed to send message {text}: {error}");
+        }
     }
 }
 
-void OnConnect()
+void OnConnect(Identity identity, string authToken)
```
RReverser, line 131: This signature doesn't match the proposal - user needs access to the DbConnection.

examples~/quickstart/client/Program.cs
```diff
@@ -88,44 +96,59 @@ void PrintMessage(Message message)
...
 
-void OnIdentityReceived(string authToken, Identity identity, Address _address)
+void OnConnectError(WebSocketError? error, string message)
 {
-    local_identity = identity;
-    AuthToken.SaveToken(authToken);
+
+}
+
+void OnDisconnect(DbConnection conn, WebSocketCloseStatus? status, WebSocketError? error)
```
RReverser, line 144: The proposal has only one error object - do we need both args here and in OnConnectError?

lcodes: Ah hmm, I think its meant to follow the idiomatic error handling in this case.

lcodes: Looking at the invocations in WebSocket.cs, there's information lost if we remove that parameter; its better to update the proposal to reflect the implementation.

examples~/quickstart/client/Program.cs
lcodes: Are we looking at the same version of the proposal? The one I have only has these two arguments.

src/SpacetimeDBClient.cs
lcodes: Ah yeah that's tricky now because the type can represent both a table with a PK and one without one, or two tables using different columns as their PKs.

Besides, the only way for updates to be collected is for their PKs to have been collected through the handle previously, which performs the check we want.

PR #132, opened 2024-09-17
Logging API

SteveGibsonCL:
## Description of Changes

Changed logging based on [this proposal](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/981)

## API

Logging interface is different now. `Logger` has been renamed to `Log`, and its methods have been renamed as well (ex. `Logger.LogError` is now `Log.Error`)

## Requires SpacetimeDB PRs


src/ConsoleLogger.cs
```diff
@@ -9,10 +9,12 @@ public enum LogLevel
...
-            Warning = 2,
-            Error = 4,
-            Exception = 8,
-            All = Debug | Warning | Error | Exception
+            Trace = 2,
+            Info = 4,
+            Warning = 8,
+            Error = 16,
+            Exception = 32,
+            All = Debug | Trace | Info | Warning | Error | Exception
```
RReverser: I know it was a hardcoded list before, but at this point maybe just

```suggestion
            All = ~0
```

RReverser (approved):
Minor nit, otherwise LGTM.

src/ConsoleLogger.cs
SteveGibsonCL: Sure

PR #136, opened 2024-09-30
Implement websocket changes atop C# sdk changes

Centril:
## Requires SpacetimeDB PRs

This is the C# side of https://github.com/clockworklabs/SpacetimeDB/pull/1761.


lcodes:
Oh good, most of the changes are to the clientAPI :)

src/SpacetimeDBClient.cs
```diff
@@ -205,6 +203,97 @@ static DbValue Decode(IRemoteTableHandle table, EncodedValue value, out object?
         private static readonly Status Committed = new Status.Committed(default);
         private static readonly Status OutOfEnergy = new Status.OutOfEnergy(default);
 
+        private static ServerMessage DecompressDecodeMessage(byte[] bytes)
+        {
+            using var stream = new MemoryStream(bytes, 1, bytes.Length - 1);
+
+            switch (bytes[0])
```
lcodes: Is it possible for bytes to be an empty array? If so it'll fail here and in creating the memory stream

src/SpacetimeDBClient.cs
```diff
@@ -205,6 +203,97 @@ static DbValue Decode(IRemoteTableHandle table, EncodedValue value, out object?
...
+            {
+                // Uncompressed.
+                case 0:
```
lcodes: Could create an enum for these, and cast the read byte to it

src/SpacetimeDBClient.cs
```diff
@@ -205,6 +203,97 @@ static DbValue Decode(IRemoteTableHandle table, EncodedValue value, out object?
...
+                    }
+                // Brotli compressed.
+                case 1:
+                    {
+                        using var decompressedStream = new BrotliStream(stream, CompressionMode.Decompress);
+                        using var binaryReader = new BinaryReader(decompressedStream);
+                        return new ServerMessage.BSATN().Read(binaryReader);
+                    }
+                default:
+                    throw new InvalidOperationException();
```
lcodes: Can we have a reason argument to the exception? "Unknown compression type" or similar

lcodes: Same for the other exceptions

tests~/SnapshotTests.VerifyAllTablesParsed.verified.txt
```diff
@@ -488,4 +486,4 @@
       Max: type=InitialSubscription
     }
   }
-}
\ No newline at end of file
+}
```
lcodes, line 393: What's this diff?

src/SpacetimeDBClient.cs
Centril: It's not possible. I'll add a comment.

tests~/SnapshotTests.VerifyAllTablesParsed.verified.txt
Centril: I'm not sure :/

lcodes (approved):

PR #139, opened 2024-10-01
Merge table cache into table handle

RReverser:
## Description of Changes

Merges cache into the table handle as suggested on the original PR + hides most table methods that shouldn't be part of the stable API.

Few remaining methods will need a codegen change to be available only to subclasses, so for now that's out of scope.

Same for merging ClientCache into RemoteTables - we shouldn't need a separate collection, and instead could autogenerate a switch expression over table name.

## API

## Requires SpacetimeDB PRs

### DIFF
--- a/src/ClientCache.cs
+++ b/src/ClientCache.cs
@@ -6,74 +6,22 @@ using SpacetimeDB.Internal;
 
 namespace SpacetimeDB
 {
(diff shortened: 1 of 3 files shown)


cloutiertyler (approved):
LGTM assuming that BitCraft doesn't use these

PR #140, opened 2024-10-01
Hide more APIs

RReverser:
## Description of Changes

Removing unstable APIs that are not used by BitCraft; marking others with [Obsolete] and renaming few others to match the proposal.

One exception is InternalCallReducer - it would need some further changes to codegen; marking it as Obsolete right now would cause all generated clients to show noisy warnings.

## API

## Requires SpacetimeDB PRs


cloutiertyler (approved):
LGTM

PR #141, opened 2024-10-02
Fix UnityNetworkManager

### LINKED ISSUE #134: UnityNetworkManager is not available before codegen
When starting a new project you're unable to add the `UnityNetworkManager` script to a GameObject in the scene because it only exists after codegen. This is an issue for 2 reasons:
1. New users who import the SpacetimeDB SDK will likely try to just immediately add this script to the scene (which is common with other network libraries like Mirror, Photon, etc.) and won't be able to find it.
2. IMO it will make the tutorial slightly less straightforward. Right now we ask the user to create a new GameObject and attach the `UnityNetworkManager` right after we import the SDK which I think makes more sense than having to wait after generation.

I think this change is important in terms of cutting down the amount of new user support we'll have to do and in general will also just help people get through the new user funnel more easily.

RReverser:
This got broken when we switched to a new DbConnection API.

Keep track of all the active connections and update/destroy them from a singleton game object. Fixes #134.

Marked as a draft for now, because it's untested and because we're not yet sure that singleton for all connections as requested is the right approach.

## Description of Changes

## API

## Requires SpacetimeDB PRs


jdetter (approved):
I tested this myself and it works. My only nit here is that we *really* should rename this to something other than `UnityNetworkManager`. My recommendation would be `SpacetimeDBNetworkManager` even though it's a bit long.

PR #143, opened 2024-10-02
Fix UnityDebugLogger implementation

RReverser:
## Description of Changes

Without explicit reference these result in

> error CS0119: 'UnityDebugLogger.Debug(string)' is a method, which is not valid in the given context

## API

## Requires SpacetimeDB PRs


bfops (approved):
Seems right to me :+1:

PR #144, opened 2024-10-02
Don't actually try to connect to network in tests

RReverser:
## Description of Changes

## API

## Requires SpacetimeDB PRs

### DIFF
--- a/src/SpacetimeDBClient.cs
+++ b/src/SpacetimeDBClient.cs
@@ -11,9 +11,6 @@ using SpacetimeDB.BSATN;
 using SpacetimeDB.Internal;
 using SpacetimeDB.ClientApi;
 using Thread = System.Threading.Thread;
-using System.Runtime.CompilerServices;
-
-[assembly: InternalsVisibleTo("SpacetimeDB.Tests")]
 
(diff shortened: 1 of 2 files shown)


cloutiertyler (approved):
LGTM

src/SpacetimeDBClient.cs
```diff
@@ -180,6 +177,7 @@ struct PreProcessedMessage
         private readonly BlockingCollection<PreProcessedMessage> _preProcessedNetworkMessages =
             new(new ConcurrentQueue<PreProcessedMessage>());
 
+        internal static bool IsTesting;
```
lcodes, line 180: Instead of the static IsTesting, applying to all connections, could we add a BuildWithoutConnect() method to the builder? seems a smaller change surface too. I'm sure we're not the only ones who'll do this sort of testing, so the builder seems like the best place for this.

RReverser: I think it's good for tests to get as far as possible to realistic scenario, including the log above (which is part of the snapshot).

But yeah, it could also be a separate method. We can revise easily in the future since the current approach is at least not part of public API so won't break others.

PR #149, opened 2024-10-03
Remove obsolete tools folder

RReverser:
## Description of Changes

Not sure when or how this was added (it's not on master), but this folder shouldn't be here - we have `tools~` instead.

## API

## Requires SpacetimeDB PRs


bfops:
This was added by https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/119, but not in the PR that cherrypicked it to `master` (https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/125).

bfops (approved):

PR #153, opened 2024-10-03
Accept multiple queries in Subscribe

RReverser:
## Description of Changes

Turns out, we're not ready for single Subscribe per query, so bringing back this ability.

## API

## Requires SpacetimeDB PRs


jdetter (approved):
I tested this myself and it is working. Thank you Ingvar 👍

PR #155, opened 2024-10-03
 Add gzip + none compression algos and let SDK pick compression

Centril:
## Description of Changes

Companion to https://github.com/clockworklabs/SpacetimeDB/pull/1802.

## Requires SpacetimeDB PRs

https://github.com/clockworklabs/SpacetimeDB/pull/1802

### DIFF
--- a/src/SpacetimeDBClient.cs
+++ b/src/SpacetimeDBClient.cs
@@ -22,6 +22,7 @@ namespace SpacetimeDB
         string? uri;
         string? nameOrAddress;
         string? token;
+        Compression? compression;
 
         public DbConnection Build()
         {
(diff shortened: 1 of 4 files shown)


src/SpacetimeDBClient.cs
```diff
@@ -205,6 +212,17 @@ enum CompressionAlgos : byte
         {
             None = 0,
             Brotli = 1,
+            Gzip = 2,
+        }
+
+        private static BinaryReader BrotliReader(Stream stream)
+        {
+            return new BinaryReader(new BrotliStream(stream, CompressionMode.Decompress));
```
lcodes, line 224: I think you could change the return signature to `StreamReader`, return the `BrotliStream` (or `GZipStream` below), and wrap the `BinaryReader` at the caller; then you can there use the switch to get the streamreader, outside the switch wrap it into binary reader and call bsatn read :)

Centril: I tried this in a commit but apparently C# doesn't let you use `return` in `switch` *expressions* so then doing the mutable dance doesn't seem worth it or even better and this seems more of a refactoring concern so I'll leave that for later.

lcodes (approved):

PR #161, opened 2024-10-04
0.12.0 DLLs (again)

bfops:
## Description of Changes
Update with the latest DLLs from the current 0.12.0 branch.

## API

## Requires SpacetimeDB PRs


jdetter (approved):
Tested, working

PR #166, opened 2024-10-12
Fix Connection Error: Success

jdetter:
## Description of Changes
This PR edits the handling of errors related to websocket connections and disconnections. In particular, clients and users would often run into the dreaded `Connection Error: Success` message which was confusing and frustrating. This PR better addresses the error by providing more guidance and debug info for the user. It is unfortunately still suboptimal because the `HttpStatusCode` is not available in the .NET core version that Unity supports. We try to be as helpful as possible in this scenario.

## API

 - [x] This is an API breaking change to the SDK, because it changes the returned values from the `OnDisconnect` and `OnConnectError` callbacks to implement the API specification: https://github.com/clockworklabs/SpacetimeDBPrivate/pull/866/files#diff-be533cc04817c33605a68d717c6ec320c4449904266ee8e1096971e9e17e8d31R424

## Requires SpacetimeDB PRs
No changes to SpacetimeDB required.

## Testing

I, Tyler, have tested this and confirmed it to be working with CircleGame. Here is a sample of the output in the case of `Connection Error: Success`:

image


cloutiertyler:
Put it to rest, once and for all!

src/WebSocket.cs
```diff
@@ -70,23 +71,81 @@ public async Task Connect(string? auth, string host, string nameOrAddress, Addre
...
+            catch (WebSocketException ex) when (ex.WebSocketErrorCode == WebSocketError.Success)
+            {
+                // How can we get here:
+                // - When you go to connect and the server isn't running (port closed) - target machine actively refused
+                // - 404 - No module with at that module address instead of 101 upgrade
+                // - 401? - When the identity received by SpacetimeDB wasn't signed by its signing key
+                // - 400 - When the auth is malformed
+                if (OnConnectError != null)
+                {
+                    // .net 6,7,8 has support for Ws.HttpStatusCode as long as you set
```
lcodes, line 100: Could we have a #ifdef here with support for .net 6+ ?

src/SpacetimeDBClient.cs
```diff
@@ -64,15 +64,15 @@ public DbConnectionBuilder<DbConnection, Reducer> OnConnect(Action<DbConnection,
             return this;
         }
 
-        public DbConnectionBuilder<DbConnection, Reducer> OnConnectError(Action<WebSocketError?, string> cb)
+        public DbConnectionBuilder<DbConnection, Reducer> OnConnectError(Action<Exception> cb)
```
lcodes, line 67: It seems now we're only using Exceptions as string wrappers. Is this because we're introducing our own error message now?

I'd be in favor of keeping the error types coming from the dotnet library in the event signature, and adding our own message on the side.

src/SpacetimeDBClient.cs
```diff
@@ -64,15 +64,15 @@ public DbConnectionBuilder<DbConnection, Reducer> OnConnect(Action<DbConnection,
...
         {
-            conn.webSocket.OnConnectError += (a, b) => cb.Invoke(a, b);
+            conn.webSocket.OnConnectError += (e) => cb.Invoke(e);
             return this;
         }
 
-        public DbConnectionBuilder<DbConnection, Reducer> OnDisconnect(Action<DbConnection, WebSocketCloseStatus?, WebSocketError?> cb)
+        public DbConnectionBuilder<DbConnection, Reducer> OnDisconnect(Action<DbConnection, Exception?> cb)
```
lcodes, line 73: Same here, how about adding a string to the event signature?

Because these callbacks serve two purposes; offer an error message, and offer a dispatch mechanism. With Exception wrapping string, the latter is lost, no?

src/WebSocket.cs
```diff
@@ -102,7 +161,45 @@ public async Task Connect(string? auth, string host, string nameOrAddress, Addre
...
-                        if (OnClose != null) dispatchQueue.Enqueue(() => OnClose(receiveResult.CloseStatus, null));
+                        if (OnClose != null)
+                        {
+                            switch (receiveResult.CloseStatus)
+                            {
+                                case WebSocketCloseStatus.NormalClosure:
+                                    dispatchQueue.Enqueue(() => OnClose(null));
+                                    break;
+                                case WebSocketCloseStatus.EndpointUnavailable:
+                                    dispatchQueue.Enqueue(() => OnClose(new Exception("(1000) The connection has closed after the request was fulfilled.")));
```
lcodes, line 172: Is there a reason why the enum is replaced by an exception wrapping a string?

src/WebSocket.cs
cloutiertyler: This is a good idea, but I’d like to do it in a separate PR since the current PR is already an improvement for the customers we have

src/SpacetimeDBClient.cs
cloutiertyler: I updated this API to match @gefjon ‘s API stability proposal. Prior to this PR it was incorrectly implementing the proposal. We’re starting broad before we expose any implementation details in our API

src/SpacetimeDBClient.cs
cloutiertyler: Same as above

src/WebSocket.cs
cloutiertyler: Same as above

lcodes (approved):

PR #168, opened 2024-10-12
Fix Reconnection Logic

jdetter:
## Description of Changes

- switched our "already connected" logic to using a reference to a `MonoBehaviour` instead of just a bool. `MonoBehaviour`s are typically destroyed when a scene reload happens and in this case we'll want to allow developers to spawn a new `SpacetimeDBNetworkManager` if the previous one has been destroyed.

## API

This is *not* an API break.

## Requires SpacetimeDB PRs

- https://github.com/clockworklabs/SpacetimeDB/pull/1869

## Testsuite

## Testing

- [x] I have added in several new tests here, one of which is a reconnection test. Also, the reason why we couldn't have more than 1 test before this is that it was required for reconnections to be working in order to have multiple tests running in the testsuite. Now that we have fixed reconnections I have enabled all of the tests.

Testsuite passes

image


jdetter:
~~This only fixes some surface level issues, there are some deeper problems here with how we handle client side indexes that will need to be fixed.~~

^ This has been fixed by Jeremie's SpacetimeDB PR

src/SpacetimeDBNetworkManager.cs
```diff
@@ -1,38 +1,53 @@
...
 using UnityEngine;
 
 namespace SpacetimeDB
 {
 	// This class is only used in Unity projects.
-	// Attach this to a gameobject in your scene to use SpacetimeDB.
+	// Attach this to a GameObject in your scene to use SpacetimeDB.
 	public class SpacetimeDBNetworkManager : MonoBehaviour
 	{
-		private static bool _alreadyInitialized;
```
jdetter, line 13: This wasn't exposed before to the users so this is *not* an API break

src/SpacetimeDBNetworkManager.cs
```diff
@@ -1,43 +1,54 @@
 #if UNITY_5_3_OR_NEWER
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using SpacetimeDB;
 using UnityEngine;
 
 namespace SpacetimeDB
 {
-	// This class is only used in Unity projects.
```
jdetter, line 9: sorry for the whitespace changes here, this file wasn't properly formatted before.

jdetter: reminder that you can view the diff with whitespace turned off
image

unity-tests/client/Assets/Scripts/autogen/_Globals/SpacetimeDBClient.cs
```diff
@@ -15,7 +15,7 @@ public sealed class RemoteTables
 	{
 		public class CircleHandle : RemoteTableHandle<EventContext, Circle>
```
jdetter: this file is automatically generated

src/SpacetimeDBNetworkManager.cs
```diff
@@ -6,38 +6,48 @@
...
-		}
+        private List<IDbConnection> cache = new List<IDbConnection>();
 
-		private void Update() => ForEachConnection(conn => conn.FrameTick());
-		private void OnDestroy() => ForEachConnection(conn => conn.Disconnect());
-	}
+        private void ForEachConnection(Action<IDbConnection> action)
+        {
+            // TODO(jdetter): We're doing this for now because we can't break the API during a minor release but
+            // in the future we should just change ActiveConnections to be a list so that we can reverse-iterate
```
RReverser: ActiveConnections is already not part of the public API though? Or do you mean some other API?

unity-tests~/client/Assets/Scripts/autogen/_Globals/SpacetimeDBClient.cs
```diff
@@ -268,15 +286,18 @@ public override void InternalInvokeValueDeleted(IDatabaseRow row)
 
 			public readonly ref struct IdentityUniqueIndex
 			{
+				readonly LoggedOutPlayerHandle Handle;
```
RReverser: The logic seems unnecessarily complicated.

Table handle stores the Dictionary<> caches.
Table handles also store instances of index structs like this IdentityUniqueIndex.
Then, index structs store a reference back to the parent table handles.
And then, index structs only access said relevant cache via those table handles.

Instead, let's move the Dictionary<> cache directly into the index struct. This way we'll encapsulate the cache data inside the relevant index struct, and won't need this kind of 3-way access just to retrieve values from the cache from inside the index.

jdetter: We moved this conversation here: https://github.com/clockworklabs/SpacetimeDB/pull/186

src/SpacetimeDBNetworkManager.cs
jdetter: I *thought* that ActiveConnections was referenced in the bindings somewhere and I was worried about creating an API break there, but I grep'd in the SpacetimeDB project for `ActiveConnections` and there's nothing there so maybe you're right. Thanks for the heads up 👍

RReverser: It has `internal` visibility so bindings couldn't use it even if they wanted to.

.npmignore
```diff
@@ -0,0 +1 @@
+unity-tests~/
```
lcodes: Is this still useful? Seems like if the module is checked out this rule come into play.

src/SpacetimeDBNetworkManager.cs
```diff
@@ -6,38 +6,48 @@
...
+                throw new InvalidOperationException("SpacetimeDBNetworkManager is a singleton and should only be attached once.");
+            }
+            else
+            {
+                _instance = this;
+            }
+        }
 
-		internal static HashSet<IDbConnection> ActiveConnections = new();
+        internal static HashSet<IDbConnection> ActiveConnections = new();
```
lcodes: Does this need to be static?

src/SpacetimeDBNetworkManager.cs
```diff
@@ -6,38 +6,48 @@
...
 
-		private void ForEachConnection(Action<IDbConnection> action)
-		{
-			foreach (var conn in ActiveConnections)
-			{
-				action(conn);
-			}
-		}
+        private List<IDbConnection> cache = new List<IDbConnection>();
```
lcodes: I don't think it's called often enough to justify a cache; how about creating this on each call to ForEach?

jdetter: `ForEach` is called every time `Update()` is called so at least once a frame. 

Ideally we wouldn't need this at all and we could just use a `List<>` instead of a `HashMap<>` but that would require changes to the bindings which is going too deep for this PR.

lcodes (approved):

PR #170, opened 2024-10-15
Companion to SpacetimeDB#1812 (light tx, C#)

Centril:
## Description of Changes

Adds C# sdk support for https://github.com/clockworklabs/SpacetimeDB/pull/1812.

## Requires SpacetimeDB PRs

- https://github.com/clockworklabs/SpacetimeDB/pull/1812

## Testsuite

SpacetimeDB branch name: centril/websocket-light


src/SpacetimeDBClient.cs
```diff
@@ -65,6 +66,12 @@ public DbConnectionBuilder<DbConnection, Reducer> WithCompression(Compression co
             return this;
         }
 
+        public DbConnectionBuilder<DbConnection, Reducer> WithLightMode(bool light)
```
gefjon, line 72: I question why this needs to take a bool, tbh, since `false` is the default and so the only reason to call this is to assign `true`, but I also don't care.

gefjon (approved):

cloutiertyler:
As far as I can determine these are missing meta files. I added the missing meta files

src/Compression.cs.meta
```diff
@@ -1,11 +1,11 @@
 fileFormatVersion: 2
-guid: 03fdc211a87474eeba9f03495eab491e
+guid: 78dd2c1276a6e4e859d2c4fad671c2c7
```
lcodes, line 2: Is it normal this guid changes?

Centril: Tyler regenerated the meta file.

jdetter: why is this file here? It seems like there is no `Compression.cs`?

Centril: There is, https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/blob/3581ed7484d3684f0d318387c487fd6b9f27fd23/src/Compression.cs

PR #182, opened 2024-10-30
Fixed subscription updates not clearing tables with no subscribed values

SteveGibsonCL:
## Description of Changes

Context: https://discord.com/channels/568217153853980682/669989878955638785/1301132060878049332
Currently, when we receive subscription updates, a table will only be diffed if subscription has any rows for that table. If, however, there are no subscribed values, that table will NOT be diffed, and therefore will not get cleared. Values from previous subscription will still be there, so the table is in incorrect state. 

This PR fixes that by making sure that ALL tables are checked in state diff

## API

## Requires SpacetimeDB PRs

## Testing

- [ ] Create a table A with an `val: i32` field
- [ ] Generate rows for table A with field values in the range (0..100)
- [ ] Connect client and subscribe to `SELECT * FROM A WHERE val > 0` (will have all rows from table A)
- [ ] Change your subscription to `SELECT * FROM A WHERE val > 1000` (should have no rows)
- [ ] Note that after subscription is applied, table A will have no values

### DIFF
--- a/src/SpacetimeDBClient.cs
+++ b/src/SpacetimeDBClient.cs
@@ -547,7 +547,10 @@ namespace SpacetimeDB
                 {
                     if (!subscriptionInserts.TryGetValue(table.ClientTableType, out var hashSet))
                     {
-                        continue;
+                        // We still need to process tables that weren't included in subscription.
+                        // Otherwise we won't delete no-longer-available values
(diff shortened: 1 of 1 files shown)


RReverser (approved):
Seems reasonable.

PR #185, opened 2024-10-30
Update the quickstart module

lcodes:
## Description of Changes

Fix the quickstart client program

## Testing

Second subscription resets the first one without the fix

## Testsuite


bfops:
Do you know _why_ this change is required? Like does this work around an underlying bug that we should be fixing in the SDK, or is this the expected usage now? (And should we be printing a message if users do trigger this by... calling `SubscriptionBuilder` more than once or something?)

lcodes:
It's the 0.12 usage until the 1.0 subscriptions are implemented. But since I keep getting burned by it when testing after switching branches, might as well solve it :)

cloutiertyler (approved):
This LGTM. For context this is currently overwriting the subscription. This will eventually be allowable as multiple subscriptions, but it is not for now.

jdetter:
@lcodes When I build the module for the quickstart I get some errors, I believe it's because we haven't changed the Cargo.toml to point to 0.12, did you run into this when you tested your PR?

```
error[E0432]: unresolved import `spacetimedb::Table`
 --> src/lib.rs:1:45
  |
1 | use spacetimedb::{ReducerContext, Identity, Table, Timestamp};
  |                                             ^^^^^ no `Table` in the root

error[E0433]: failed to resolve: could not find `table` in `spacetimedb`
 --> src/lib.rs:4:16
  |
4 | #[spacetimedb::table(name = user, public)]
  |                ^^^^^ could not find `table` in `spacetimedb`

error[E0433]: failed to resolve: could not find `table` in `spacetimedb`
  --> src/lib.rs:12:16
   |
12 | #[spacetimedb::table(name = message, public)]
   |                ^^^^^ could not find `table` in `spacetimedb`

error[E0433]: failed to resolve: could not find `reducer` in `spacetimedb`
  --> src/lib.rs:19:16
   |
19 | #[spacetimedb::reducer(init)]
   |                ^^^^^^^ could not find `reducer` in `spacetimedb`
```

examples~/quickstart/server/Cargo.lock
```diff
@@ -62,6 +62,18 @@ dependencies = [
  "generic-array",
 ]
 
+[[package]]
```
jdetter: Did we mean to commit this?

PR #193, opened 2024-11-25
Add expiration for stored request durations

RReverser:
## Description of Changes

This adds configurable expiration time for request durations. This is to prevent memory leak in long-running applications where otherwise the queue would only keep growing indefinitely.

## API

## Requires SpacetimeDB PRs

## Testsuite
*If you would like to run against a specific SpacetimeDB branch in the testsuite, specify that here. This can be a branch name or a link to a PR.*


bfops:
Migrated to https://github.com/clockworklabs/SpacetimeDB/pull/3148, as we are merging this repo into that one.

PR #211, opened 2025-01-10
`SubscribeToAllTables`, which hides `"SELECT * FROM *"`

gefjon:
## Description of Changes

Per out-of-band discussion, add `SubscriptionBuilder.SubscribeToAllTables`, which abstracts over `SubscriptionBuilder.Subscribe(["SELECT * FROM *"])`.

In the future, we will change the implementation of this method, so that it uses "legacy subscriptions" while `SubscriptionBuilder.Subscribe` moves to using "mutable subscriptions." At that time, no other interface will be provided for using "legacy subscriptions." `SubscribeToAllTables` may also at some point be rewritten in terms of "mutable subscriptions" somehow.

## API

## Requires SpacetimeDB PRs

N/a

## Testsuite

## Testing

- [x] @cloutiertyler will use this in the new tutorial and report back.


cloutiertyler (approved):
The change is good. I will investigate the failing tests and see if we can't fix those.

bfops:
The CI here should be fixed once https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/210 merges.

That PR should be merged once we've pushed NuGet versions of the BSATN DLLs.

src/Event.cs
```diff
@@ -62,6 +62,11 @@ public SubscriptionBuilder<EventContext> OnError(Action<EventContext> callback)
         }
 
         public SubscriptionHandle<EventContext> Subscribe(params string[] querySqls) => new(conn, Applied, Error, querySqls);
+
+        public void SubscribeToAllTables() {
```
RReverser: Hm wouldn't it be better to return the subcription handle from the API to give user some control if they need it?

src/Event.cs
```diff
@@ -62,6 +62,11 @@ public SubscriptionBuilder<EventContext> OnError(Action<EventContext> callback)
...
+            string[] querySqls = {"SELECT * FROM *"};
+            new SubscriptionHandle<EventContext>(conn, Applied, Error, querySqls);
```
RReverser: We accept `params` in the API, which allows variadic params as a syntax sugar instead of an array, so you can just do

```suggestion
            new SubscriptionHandle<EventContext>(conn, Applied, Error, "SELECT * FROM *");
```

src/Event.cs
gefjon: No, we explicitly want to give the user as little control as possible if they use this method. It's for simple apps where you don't care; if you want/need more control you should be writing out queries.

When we implement the new subscription API, we're going to keep this method in terms of "legacy subscriptions" (i.e. the current API). This will likely not play well with the `SubscriptionHandle` machinery.

src/Event.cs
```diff
@@ -62,6 +62,10 @@ public SubscriptionBuilder<EventContext> OnError(Action<EventContext> callback)
         }
 
         public SubscriptionHandle<EventContext> Subscribe(params string[] querySqls) => new(conn, Applied, Error, querySqls);
+
+        public void SubscribeToAllTables() {
+            new SubscriptionHandle<EventContext>(conn, Applied, Error, "SELECT * FROM *");
```
RReverser: Oops, sorry, it's the `Subscribe` function that has variadic signature, not the constructor, so this should actually be just

```suggestion
            Subscribe("SELECT * FROM *");
```

RReverser (approved):

bfops:
CI now failing due to whitespace errors

cloutiertyler:
Although the cost was high (see #217) I have verified that this change is working appropriately in the tutorial.

PR #212, opened 2025-01-13
Revise `WithCredentials` to `WithToken`

gefjon:
## Description of Changes

`WithCredentials` was designed for our old auth model, where tokens were always issued by the SpacetimeDB host alongside an `Identity`, and therefore the bearer always knew their `Identity`.

In our new auth model, a client may have a valid auth-able JWT but not know what `Identity` it will result in,
as the `Identity` is computed based on the hash of some of the token's contents, and this hashing algorithm is not included in our client SDKs. As such, this PR revises `WithCredentials` to `WithToken`, which just accepts the token.

## API

`WithCredentials` is renamed and its signature changes.

## Requires SpacetimeDB PRs

## Testsuite

## Testing

- [ ] @cloutiertyler should test this with the Unity tutorial, if possible.


RReverser (approved):
Looks pretty straightforward.

PR #215, opened 2025-01-14
Unify CI handling of SpacetimeDB branch

RReverser:
## Description of Changes

This unifies various CI task into a single config that checks out the referenced SpacetimeDB branch only once and using the same PR parsing code (previously used only in Unity), and then runs all the builds and tests against it. That is, .NET test, server module and Unity tests are now all consistently overridden to point to the same SpacetimeDB.

## API

## Requires SpacetimeDB PRs

## Testsuite

## Testing

- [x] Lots of CI testing and tweaking until all of existing CI tests started working again; ensured .NET tests are executed as expected against the specified branch name.


.github/workflows/test.yml
```diff
@@ -0,0 +1,150 @@
...
+          # Write out the nuget config file to `nuget.config`. This causes the spacetimedb-csharp-sdk repository
+          # to be aware of the local versions of the `bindings-csharp` packages in SpacetimeDB, and use them if
+          # available. Otherwise, `spacetimedb-csharp-sdk` will use the NuGet versions of the packages.
+          # This means that (if version numbers match) we will test the local versions of the C# packages, even
+          # if they're not pushed to NuGet.
+          # See https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file for more info on the config file.
+          ./tools~/write-nuget-config.sh SpacetimeDB~
+
+      - name: Run .NET tests
+        run: dotnet test -warnaserror
```
cloutiertyler, line 68: NIT: How long to the dotnet tests take to run? In principle, this runs something in serial that was previously run in parallel, although if this doesn't take much time it doesn't really matter.

RReverser: They are very quick compared to Unity tests, like less than 10 sec or so, so I figured it's fine.

cloutiertyler (approved):
I've gone through the old workflows and the new one and visually confirmed that all the old stuff is in the new one. I think this is an improvement on the old one since there's not nearly as much duplicated work and code which could lead to inconsistencies between the tests.

RReverser:
Looks like this needs fixups after those binary splits / renames. I'll fix it tomorrow.

The other thing I'm surprised about is that the PR status checks show 2 pending checks for CI actions that no longer exist after this PR. I wonder if it's because those checks are marked as required in repo settings and they need to be unmarked for PR to become mergeable?

cloutiertyler:
Yes, I'll just have to fix that up in project settings right before we merge. Don't want to disable them prematurely because they should still apply to other PRs.

PR #216, opened 2025-01-14
Disable PR comment for meta file check if not a PR

RReverser:
## Description of Changes

Looks like the .meta check hard-fails on push events if `enable_pr_comment` is `true`, instead of just ignoring it...

I hope this fix works, but hard to tell without merging first.

## API

## Requires SpacetimeDB PRs

## Testsuite

## Testing

- [x] This is a PR that changes CI configuration, so see CI


bfops (approved):
Sure, seems fine.

PR #219, opened 2025-01-17
Implement new subscription API

kazimuth:
## Description of Changes

As proposed. No upstream codegen changes needed :)

## API

The subscription API is slightly different.

## Requires SpacetimeDB PRs
https://github.com/clockworklabs/SpacetimeDB/pull/2111

## Testsuite

SpacetimeDB branch name: jsdt/subscribe-sdk-3

## Testing
So far I have performed manual testing with the chat example. Working on updating the unity and unit tests.


examples~/quickstart/client/Program.cs
```diff
@@ -1,19 +1,21 @@
...
 
 const string HOST = "http://localhost:3000";
 const string DBNAME = "chatqs";
 
 // our local client SpacetimeDB identity
-Identity? local_identity = null;
+Identity? localIdentity = null;
 // declare a thread safe queue to store commands
-var input_queue = new ConcurrentQueue<(string Command, string Args)>();
+var inputQueue = new ConcurrentQueue<(string Command, string Args)>();
```
kazimuth: Ah, this shouldn't be here.

src/Event.cs
```diff
@@ -62,41 +62,118 @@ public SubscriptionBuilder<EventContext> OnError(Callback callback)
...
+
+        public bool IsEnded
+        {
+            get; private set;
+        }
+
+        public bool IsActive
+        {
+            get; private set;
+        }
```
RReverser: Nit: I wonder if these should be backed by a private enum state like `{ Initial, Active, Ended }` to avoid accidental confusion in `IsActive` / `IsEnded` assignments.

src/Event.cs
```diff
@@ -62,41 +62,118 @@ public SubscriptionBuilder<EventContext> OnError(Callback callback)
...
+        /// </summary>
+        /// <param name="conn"></param>
+        /// <param name="onApplied"></param>
+        /// <param name="onError"></param>
+        /// <param name="querySqls"></param>
         internal SubscriptionHandle(IDbConnection conn, SubscriptionBuilder<EventContext>.Callback? onApplied, SubscriptionBuilder<EventContext>.Callback? onError, string[] querySqls)
         {
+            IsActive = false;
+            IsEnded = false;
+            this.onApplied = onApplied;
```
RReverser, line 183: If this constructor doesn't assign `onError`, should the param be removed maybe?

src/Event.cs
```diff
@@ -62,41 +62,118 @@ public SubscriptionBuilder<EventContext> OnError(Callback callback)
...
+        public void Unsubscribe()
+        {
+            UnsubscribeThen(null);
+        }
+
+        /// <summary>
+        /// Unsubscribe from the query controlled by this subscription handle,
+        /// and call onEnded when its rows are removed from the client cache.
+        /// </summary>
+        public void UnsubscribeThen(SubscriptionBuilder<EventContext>.Callback? onEnded)
```
RReverser, line 217: Is implementation of this method supposed to actually unsubscribe by sending a websocket message or something?

src/SpacetimeDBClient.cs
```diff
@@ -377,30 +388,66 @@ HashSet<byte[]> GetInsertHashSet(System.Type tableType, int tableSize)
...
+                        }
+                        var obj = table.DecodeValue(bin);
+                        var op = new DbOp
+                        {
+                            table = table,
+                            insert = new(obj, bin),
+                        };
+                        dbOps.Add(op);
+                    }
+                }
```
RReverser: Looks like this entire loop is duplicated with PreProcessLegacySubscription. I know we're going to eventually remove the legacy one, but for now can it be extracted into a helper function to simplify maintainance?

src/SpacetimeDBClient.cs
```diff
@@ -954,4 +1086,22 @@ public void FrameTick()
...
+        /// </summary>
+        private int lastAllocated;
+
+        /// <summary>
+        /// Allocate a new ID in a thread-safe way.
+        /// </summary>
+        /// <returns>A previously-unused ID.</returns>
+        public uint Next()
+        {
+            var next = Interlocked.Increment(ref lastAllocated);
```
RReverser: Huh, that's a good one. I guess StartTrackingRequest should be doing the same, and might be potentially racy already (although it's less critical functionality).

Maybe worth switching its `_nextRequestId` to your `UintAllocator` as well.

src/SpacetimeDBClient.cs
```diff
@@ -954,4 +1086,22 @@ public void FrameTick()
             }
         }
     }
+
+    internal struct UintAllocator
+    {
+        /// <summary>
+        /// Interlocked.Increment only works with ints.
```
RReverser: I was surprised as I'm pretty sure it works with `uint`, but looks like not in .NET Standard. Oh, Unity... https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment?view=net-9.0#system-threading-interlocked-increment(system-uint32@)

src/SpacetimeDBClient.cs
```diff
@@ -141,8 +142,22 @@ struct DbOp
         [Obsolete]
         public event Action<Exception>? onSendError;
 
+        /// <summary>
+        /// Dictionary of legacy subscriptions, keyed by request ID rather than query ID.
+        /// Only used for `SubscribeToAllTables()`.
+        /// </summary>
+        private readonly Dictionary<uint, ISubscriptionHandle> legacySubscriptions = new();
```
RReverser, line 150: Hm, I assume you made id increment atomic to allow `.Subscribe()` from multiple threads.

I guess that's not the usecase we thought of in the beginning, but if we want to allow that, should those dictionaries also be made `ConcurrentDictionary` instead? Or, alternatively, should they be locked when doing any ops?

RReverser: Alternatively, I think it's also fine to revert to regular `int` and leave support for `.Subscribe` from multiple threads for a future PR. It's certainly not very common usecase.

kazimuth: I was just blindly copying @jsdt's implementation. Happy to simplify.

src/Event.cs
kazimuth: I was originally doing fancier logic here to deal with disconnect being called multiple times, then I gave up on that and just throw an error, this is a remnant of that. Will simplify.

kazimuth:
Added https://github.com/clockworklabs/Blackholio/pull/13 with testing updates. I was able to run Blackholio locally with multiple connections. I guess I have some fix I need to do on that PR to get the CI to stop complaining here.

kazimuth:
This is now ~ready for merge pending a bit of cleanup in Blackholio. 

I don't have a test that fully exercises the new disconnection behavior but I'd like to merge this as-is and add that later. Since the tutorials are designed to use SubscribeToAllTables, none of the new disconnection behavior will be observable in them. It will need a test framework that can actually do full integration tests between this SDK and master to write such tests.

kazimuth:
Because Github Actions' design is asinine, I can't update the workflow file in this branch and have the updated workflow run here. I can't use a workflow dispatch either without committing to the main branch. So, we're going to be forced to merge this with broken tests.

kazimuth:
Not ready for merge yet though, still working on the C# server.

RReverser:
If you are changing CI config, I wonder if we should merge https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/215 first.

Otherwise it seems there will be a lot of conflicts.

Plus, the deduplication that PR does might make further CI changes easier?

examples~/quickstart/client/Program.cs
```diff
@@ -1,10 +1,12 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Data;
```
cloutiertyler: Do we need this?

examples~/quickstart/client/Program.cs
```diff
@@ -135,9 +137,28 @@ void OnConnect(DbConnection conn, Identity identity, string authToken)
...
+    };
+
+    var userSubscription = conn.SubscriptionBuilder()
+        .OnApplied(waitForSubscriptions)
+        .Subscribe("SELECT * FROM user");
+    var messageSubscription = conn.SubscriptionBuilder()
+        .OnApplied(waitForSubscriptions)
+        .Subscribe("SELECT * FROM message");
+
+    //conn.SubscriptionBuilder().OnApplied(OnSubscriptionApplied).SubscribeToAllTables();
```
cloutiertyler: Can we just leave a comment saying that this is how you subscribe to all tables.

examples~/quickstart/client/Program.cs
```diff
@@ -135,9 +137,28 @@ void OnConnect(DbConnection conn, Identity identity, string authToken)
...
+    {
+        // Note: callbacks are always invoked on the main thread, so you don't need to
+        // worry about thread synchronization or anything like that.
+        subscriptions += 1;
+
+        if (subscriptions == 2)
+        {
+            OnSubscriptionApplied(ctx);
+        }
+    };
```
cloutiertyler, line 147: Sad that we have to do this, but it is what it is.

src/SpacetimeDBClient.cs
```diff
@@ -883,6 +1030,29 @@ void IDbConnection.Subscribe(ISubscriptionHandle handle, string[] querySqls)
             ));
         }
 
+        void IDbConnection.Subscribe(ISubscriptionHandle handle, string querySql)
+        {
+            if (!webSocket.IsConnected)
+            {
+                Log.Error("Cannot subscribe, not connected to server!");
```
cloutiertyler, line 1037: Shouldn't this be throwing an exception instead?

tests~/SnapshotTests.VerifySampleDump_dumpName=SubscribeApplied.verified.txt
```diff
@@ -0,0 +1,558 @@
+﻿{
+  Events: {
+    Log: SpacetimeDBClient: Connecting to wss://spacetimedb.com example,
+    OnConnect: {
+      identity: Identity_1,
+      token: eyJ0eXAiOiJKV1QiLCJhbGciOiJFUzI1NiJ9.eyJoZXhfaWRlbnRpdHkiOiI4ZjkwY2M5NGE5OTY4ZGY2ZDI5N2JhYTY2NTAzYTg5M2IxYzM0YjBiMDAyNjhhNTE0ODk4ZGQ5NTRiMGRhMjBiIiwiaWF0IjoxNzE4NDg3NjY4LCJleHAiOm51bGx9.PSn481bLRqtFwIh46nOXDY14X3GKbz8t4K4GmBmz50loU6xzeL7zDdCh1V2cmiQsoGq8Erxg0r_6b6Y5SqKoBA
```
cloutiertyler: This is an old token that is invalid now.

cloutiertyler (approved):
I have tested this in Unity by running Blackholio against it. Specifically branch: https://github.com/clockworklabs/Blackholio/pull/13

Also I reviewed the code and asked questions which have been answered by @kazimuth

src/SpacetimeDBClient.cs
cloutiertyler: TL;DR James says no in so far as he was following the existing convention.

I would like to ask @gefjon at some point if this should eventually throw an Exception.

tests~/SnapshotTests.VerifySampleDump_dumpName=SubscribeApplied.verified.txt
cloutiertyler: We updated this.

PR #224, opened 2025-01-27
Revert DLLs to `1.0.0-rc3`

bfops:
## Description of Changes
Revert DLLs, which I should have done in https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/223.

See that PR for background on why it was necessary.

## API

Kinda, but it's a revert.

## Requires SpacetimeDB PRs

## Testsuite

## Testing
I would have liked to say that the current CI is sufficient as I did in the last PR, but now I no longer feel confident.. Currently the repo is in a somewhat inconsistent state, though, so I think we should merge this right now as it will not make it _less_ consistent.

I'm very open to suggestions about how to test this properly.


RReverser (approved):

cloutiertyler (approved):
I sync'd with Zeke in a video call. We agreed that this change being a hotfix should not have merged back into staging in this case. When we release rc4, this change should be moved into master, and it will be breaking because removes the old way to do subscriptions.

bfops:
How on *earth* has this been running for 42m?

RReverser:
Might be worth restarting at this point.

PR #225, opened 2025-01-27
Bump version to `1.0.0-rc4`

bfops:
## Description of Changes
Corresponding change to https://github.com/clockworklabs/SpacetimeDB/pull/2177. See that PR for more details.

~~Note that this PR only bumps the versions, but does **not** update the DLLs. This is because SpacetimeDB will likely have further changes, so the DLLs will just need to be updated again (in principle, they should be updated every time we push to SpacetimeDB `master`, which isn't really feasible. For this reason, users should not use `staging` without having their own copy of the SpacetimeDB repo as well).~~

## API

No breaking changes.

## Requires SpacetimeDB PRs
https://github.com/clockworklabs/SpacetimeDB/pull/2177

## Testsuite

## Testing
- [x] CI passes with that branch name
- [x] The branch name has been changed back to `master` after that PR merges and CI still passes


RReverser:
> Note that this PR only bumps the versions, but does not update the DLLs. This is because SpacetimeDB will likely have further changes, so the DLLs will just need to be updated again

I think updating DLLs here is still important, because this is the PR that will change the folder name `packages/spacetimedb.bsatn.runtime/1.0.0-rc3` to `.../1.0.0-rc4`.

If SpacetimeDB master has further DLL updates, that's fine, but updating DLLs on version bumps is probably more important.

bfops:
> > Note that this PR only bumps the versions, but does not update the DLLs. This is because SpacetimeDB will likely have further changes, so the DLLs will just need to be updated again
> 
> I think updating DLLs here is still important, because this is the PR that will change the folder name `packages/spacetimedb.bsatn.runtime/1.0.0-rc3` to `.../1.0.0-rc4`.
> 
> If SpacetimeDB master has further DLL updates, that's fine, but updating DLLs on version bumps is probably more important.

Yeah that's a great point, doing that now.

RReverser (approved):

bfops:
@cloutiertyler said we can merge without his review

PR #228, opened 2025-01-28
Improve CI caching

RReverser:
## Description of Changes

Use SpacetimeDB commit hash in the cache key and skip rebuild/reinstallation altogether if we got an exact hit.

This saves 6-8 minutes off CI time on reruns.

## API

## Requires SpacetimeDB PRs

## Testsuite

## Testing

- [x] CI testing forth and back until it worked correctly.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,6 +45,7 @@ jobs:
 
       - name: Checkout SpacetimeDB
         uses: actions/checkout@v4
+        id: checkout-stdb
         with:
           repository: clockworklabs/SpacetimeDB
           ref: ${{ steps.extract-branch.outputs.branch }}
@@ -83,19 +84,23 @@ jobs:
 
       - name: Cache Rust dependencies
         uses: Swatinem/rust-cache@v2
+        id: cache-rust-deps
         with:
           workspaces: unity-tests~/server-rust
-          shared-key: UnityTestServer
+          key: ${{ steps.checkout-stdb.outputs.commit }}
+          # Cache Rust deps even if unit tests have failed.
+          cache-on-failure: true
+          # Cache the CLI as well.
+          cache-all-crates: true
 
       - name: Install SpacetimeDB CLI from the local checkout
+        # Rebuild only if we didn't get a precise cache hit.
+        if: steps.cache-rust-deps.outputs.cache-hit == 'false'
         run: |
           cargo install --force --path SpacetimeDB~/crates/cli --locked --message-format=short
           cargo install --force --path SpacetimeDB~/crates/standalone --locked --message-format=short
           # Add a handy alias using the old binary name, so that we don't have to rewrite all scripts (incl. in submodules).
-          rm $HOME/.cargo/bin/spacetime || echo "haven't run on this host before"
-          ln -s $HOME/.cargo/bin/spacetimedb-cli $HOME/.cargo/bin/spacetime
-          # Clear any existing information
-          spacetime server clear -y
+          ln -sf $HOME/.cargo/bin/spacetimedb-cli $HOME/.cargo/bin/spacetime
         env:
           # Share the target directory with our local project to avoid rebuilding same SpacetimeDB crates twice.
           CARGO_TARGET_DIR: unity-tests~/server-rust/target
@@ -129,14 +134,6 @@ jobs:
           spacetime logout && spacetime login --server-issued-login local
           bash ./publish.sh
 
-      - uses: actions/cache@v3
-        with:
-          path: unity-tests~/client/Library
-          key: Library-SpacetimeDBUnityTestsuite-Linux-x86
-          restore-keys: |
-            Library-SpacetimeDBUnityTestsuite-
-            Library-
-
       - name: Patch com.clockworklabs.spacetimedbsdk dependency in manifest.json
         working-directory: unity-tests~/client/Packages
         run: |
@@ -144,6 +141,12 @@ jobs:
           # TODO: find out why pointing to a local directory doesn't work - is it because Unity CI action uses Docker?
           yq e -i '.dependencies["com.clockworklabs.spacetimedbsdk"] = "https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk.git#${{ github.head_ref }}"' manifest.json
           cat manifest.json
+
+      - uses: actions/cache@v3
+        with:
+          path: unity-tests~/client/Library
+          key: Unity-${{ github.head_ref }}
+          restore-keys: Unity-
 
       - name: Run Unity tests
         uses: game-ci/unity-test-runner@v4


.github/workflows/test.yml
```diff
@@ -83,19 +84,21 @@ jobs:
...
         run: |
-          cargo install --force --path SpacetimeDB~/crates/cli --locked --message-format=short
-          cargo install --force --path SpacetimeDB~/crates/standalone --locked --message-format=short
-          # Add a handy alias using the old binary name, so that we don't have to rewrite all scripts (incl. in submodules).
-          rm $HOME/.cargo/bin/spacetime || echo "haven't run on this host before"
-          ln -s $HOME/.cargo/bin/spacetimedb-cli $HOME/.cargo/bin/spacetime
-          # Clear any existing information
-          spacetime server clear -y
+          cargo install --path SpacetimeDB~/crates/cli --locked --message-format=short
+          cargo install --path SpacetimeDB~/crates/standalone --locked --message-format=short
```
bfops: why do we no longer need to e.g. `ln spacetimedb-cli` or add `--force` etc?

bfops: ah I assume it's because we're now caching only for this specific commit

RReverser: `--force` is implied by `--path` so not necessary. (at first I thought it was also ignoring cached binaries, but turned out it just doesn't do anything at all)

As for `ln`, that's in response to upstream change - the binary got renamed from `spacetimedb-cli` to `spacetime` so `ln` would override the actual binary with a self-symlink.

That said, it might be just on the SpacetimeDB branch I pointed to - I couldn't point to master because https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/225 is not merged yet.

Happy to wait for that PR to be merged first, and then rebase again as a final check if that's easier.

bfops: I'm super confused. What do you mean `--path` implies `--force`? I'm pretty sure I had to add `--force` in order to get CI to pass in a recent PR o_O.

Re: ln - Afaik it's the other way around.. the split binaries PR created the quite new `spacetimedb-cli` binary. It's just that that change _isn't_ in the release tag you're pointing to (`v1.0.0-rc3-hotfix2`).

bfops: I assumed we got to get rid of this "overwrite existing stuff" behavior because we changed the cache key, no?

(That said, doesn't that mean we're going to be rebuilding more frequently?)

bfops:
I think I should probably ask a higher-level question so I'm not on the wrong train of thought - what's the goal of this change / what problem are we trying to solve with the previous behavior? (e.g. reduce rebuilds, solve a correctness issue, just simplify the action code, or something else)

.github/workflows/test.yml
bfops: From https://doc.rust-lang.org/cargo/commands/cargo-install.html:

> Installing with `--path` will always build and install, unless there are conflicting binaries from another package. The `--force` flag may be used to force Cargo to always reinstall the package.

I don't think `--path` implies `--force`?

Here's me getting an error without it:
image

RReverser: > Re: ln - Afaik it's the other way around.. the split binaries PR created the quite new `spacetimedb-cli` binary. It's just that that change _isn't_ in the release tag you're pointing to (`v1.0.0-rc3-hotfix2`).

Ah yeah that would make sense too. I just know that I had to bang my head trying to figure out why it's complaining about `spacetime` not existing, when it was just a recursive `ln` 😅 Unfortunately, I can't use any newer release branches due to version mismatch.

> I'm super confused. What do you mean `--path` implies `--force`? I'm pretty sure I had to add `--force` in order to get CI to pass in a recent PR o_O.

Uh, that shouldn't be necessary. I even tested locally by doing `cargo install --path` couple of times from the same path and it upgraded happily.

> (That said, doesn't that mean we're going to be rebuilding more frequently?)

No, rust-cache restores imprecise matches as well, and will reuse as much as possible from them; it's just that if you have a direct cache hit (matched by SpacetimeDB SHA), then we know we can skip rebuilding altogether.

RReverser:
> I think I should probably ask a higher-level question so I'm not on the wrong train of thought - what's the goal of this change / what problem are we trying to solve with the previous behavior? (e.g. reduce rebuilds, solve a correctness issue, just simplify the action code, or something else)

It's to reduce rebuilds / rebuild times.

Something I've been struggling with a lot while working on https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/220 - where I noticed that despite having SpacetimeDB branch unchanged and pointing to the same commit, CI is wasting a lot of time rebuilding SpacetimeDB CLI from scratch over and over, together with all our dependencies.

Caching was essentially useless / not used properly, which led me down this rabbit hole figuring out why and how it can be improved.

.github/workflows/test.yml
RReverser: > Here's me getting an error without it:

That's super weird... I can't repro locally. Okay, I'll add back `--force`.

RReverser: And actually your quote from the docs confirms it should work:

> Installing with --path **will always** build and install, unless there are conflicting binaries from **another package**.

That is, doing repetitive `cargo install --path` from the _same_ package should work fine.

RReverser: Ah I think in your screenshot you're getting error because it tries to overwrite the symlink we had in cache, with an actual binary. As long as it's just `cargo install` overriding binaries it produced, it works as expected.

Either way, I reverted that change. Let's get `staging` to working state again, and I'll rebase this PR then to ensure that it does work as expected (e.g. will need to revert the `ln` change).

RReverser:
This is rebased & fixed up now btw.

kazimuth (approved):
This seems extremely helpful.

bfops:
I kinda worry that we should not have removed
```
          rm $HOME/.cargo/bin/spacetime || echo "haven't run on this host before"
```
and
```
          # Clear any existing information
          spacetime server clear -y
```
Since those were probably added for real reasons, but I guess we'll find out.

RReverser:
> `rm $HOME/.cargo/bin/spacetime`

This is covered by the `f` in `ln -sf` now.

bfops:
> > `rm $HOME/.cargo/bin/spacetime`
> 
> This is covered by the `f` in `ln -sf` now.

Well, as long as `spacetime` was a symlink before. I imagine we had some troubles when we were switching between branches with different approaches to the `spacetime` binary. Hopefully we're past that now.

RReverser:
> Well, as long as `spacetime` was a symlink before.

AFAIK `-f` overrides the destination file regardless of what it is:

```
-f, --force
              remove existing destination files
```

bfops:
> > Well, as long as `spacetime` was a symlink before.
> 
> AFAIK `-f` overrides the destination file regardless of what it is:
> 
> ```
> -f, --force
>               remove existing destination files
> ```

Yup you're right :+1: idk why I thought it didn't overwrite normal files.

PR #229, opened 2025-01-29
Fix handling of multi-tables

RReverser:
## Description of Changes

Unfortunately, none of our tests currently cover this, but while working on the V9 upgrade, I noticed that this code still relies on `type(Row)` as a unique table identifier.

That no longer holds with multi-tables as several tables can share the same `Row` type. In that case, subscription updates would be grouped incorrectly and always applied to the same first table that uses `Row` for its data storage.

This PR fixes that by using the table handle itself as a key (compared by reference).

If transaction updates are already grouped uniquely by table, it should be possible to simplify this code much further, but I'm not sure if such guarantee exists, so leaving that untouched.

## API

## Requires SpacetimeDB PRs

## Testsuite

## Testing

- [x] I did `dotnet test`, but as mentioned in the description, this requires adding tests for multi-table subscriptions, which I'm afraid I won't have time to do, so have to leave to follow-up devs.


kazimuth (approved):
Already looked at this a few days ago and forgot to review, but this makes perfect sense and I'm happy with it.

kazimuth:
Have also tested this locally with blackholio, merging.

PR #240, opened 2025-02-07
Rework event contexts in C#

kazimuth:
## Description of Changes
https://github.com/clockworklabs/SpacetimeDB/pull/2226

## API

callbacks are better

## Requires SpacetimeDB PRs
https://github.com/clockworklabs/SpacetimeDB/pull/2226

## Testsuite

SpacetimeDB branch name: jgilles/csharp-better-events

## Testing

- [x] quickstart
- [x] blackholio

### DIFF
--- a/src/Event.cs
+++ b/src/Event.cs
@@ -38,38 +38,49 @@ namespace SpacetimeDB
     }
 
     // TODO: Move those classes into EventContext, so that we wouldn't need repetitive generics.
-    public sealed class SubscriptionBuilder<EventContext>
-        where EventContext : IEventContext
+    public sealed class SubscriptionBuilder<SubscriptionEventContext, ErrorContext>
+        where SubscriptionEventContext : IEventContext
+        where ErrorContext : IEventContext
     {
         private readonly IDbConnection conn;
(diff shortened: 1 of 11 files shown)


src/Event.cs
```diff
@@ -3,7 +3,27 @@
...
+    }
+
+    public interface ISubscriptionEventContext
+    {
+
+    }
+
+    public interface IErrorContext
+    {
+        public Exception Error { get; }
```
gefjon: Can we remove this? The error should already be in `ctx.Event`, and also exposed as a callback arg.

gefjon (approved):
One request :pray: , but then this looks good. Thanks so much for the quick work in this trying time!

src/Event.cs
kazimuth: Need it to move the error to the callback arg

kazimuth: I guess I can just call it Event

gefjon: All of these types (except `SubscriptionEventContext`) should have an `Event` field, right?

kazimuth: It's there in the generated code. This interface pattern is harebrained but I don't want to change it rn

PR #249, opened 2025-02-14
SubscribeMulti + UnsubscribeMulti, multiplicities

kazimuth:
## Description of Changes
Add SubscribeMulti and UnsubscribeMulti from upstream.
Fix unsubscribe bug found by testing against Bitcraft: Calling `UnsubscribeThen` with any (non-null) callback would result in an exception incorrectly telling the user that `Unsubscribe` had been called twice.
Multiplicity support is implemented with unit tests + manual testing of quickstart-chat.
Also, rows in the client cache are now looked up by primary key if available, which I suspect is going to be a large performance boost.

## API

## Requires SpacetimeDB PRs

## Testsuite

SpacetimeDB branch name: jgilles/final-cs-codegen-changes

## Testing

- [x] unit tests
- [ ] Working with Ryan to add a test suite.
- [x] blackholio


bfops:
I changed the base branch to `master`; I'll re-cut the release branch when things settle down.

bfops:
Out of curiosity, why did the DLLs need updating?

src/Event.cs
```diff
@@ -254,7 +254,7 @@ public void UnsubscribeThen(Action<SubscriptionEventContext>? onEnded)
             {
                 throw new Exception("Cannot unsubscribe from inactive subscription.");
             }
-            if (onEnded != null)
+            if (this.onEnded != null)
```
bfops, line 278: it also looks like we have a separate "bug" where calling `Unsubscribe` twice won't actually trigger the exception below? (Because `this.onEnded` will be `null` if we called `Unsubscribe` since that triggers this function with a `null` param) Is that.. an issue..?

kazimuth: GOOOD point

kazimuth:
Blackholio tested on https://github.com/clockworklabs/Blackholio/pull/22

kazimuth:
@bfops uhh what branch should this be based on?

joshua-spacetime:
@kazimuth if I'm reading this correctly, inserts are applied to the cache before deletes, correct?

kazimuth:
Good catch -- but that loop is accumulating into a MultiDictionaryDelta, which is a collection of key-value pairs with signed multiplicities. Since the multiplicities can go negative, you're allowed to remove a row before you insert it; in general, the order Adds and Removes are applied to a MultiDictionaryDelta doesn't matter. The resulting delta will have the same effect when Applied to a MultiDictionary. The proptest ShuffleDelta tests this.

I should definitely comment about this, thank you.

src/MultiDictionary.cs
```diff
@@ -0,0 +1,416 @@
...
+    /// A bulk change to a multidictionary. Allows both adding and removing rows.
+    /// 
+    /// Can be applied to a multidictionary, and also inspected before application to see
+    /// what rows will be deleted. (This is used for OnBeforeDelete.)
+    /// 
+    /// Curiously, the order of operations applied to a MultiDictionaryDelta does not matter.
+    /// No matter the order of Add and Remove calls on a delta, when the Delta is applied,
+    /// the result will be the same, as long as the Add and Remove *counts* for each KeyValuePair are
+    /// the same.
+    /// (This means that this is a "conflict-free replicated data type", unlike MultiDictionary.)
```
kazimuth, line 303: @joshua-spacetime

kazimuth:
@bfops , possibly the DLLs did NOT need updating and I did it unnecessarily. I just ran my update-everything script out of habit, but I can undo that.

Update: Done.

examples~/multiplicity/client/Program.cs
```diff
@@ -0,0 +1,526 @@
+﻿// Multiplicity Test Client
+// This test adds several dogs to the Multiplicity Test Server,
+// stores a local model of those dogs,
+// creates several connections in different combinations,
```
kazimuth: Thanks for the test @rekhoff !

```suggestion
// creates several subscriptions in different combinations,
```

examples~/multiplicity/client/nuget.config
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
```
kazimuth: Ah, this file should be .gitignored.

kazimuth:
@rekhoff I'm wondering if we should put that multiplicity test into its own PR. I think we probably want to merge this and add that test later. It will require some git-fu I can help out with.

examples~/multiplicity/client/Model.cs
```diff
@@ -0,0 +1,164 @@
...
+
+namespace client;
+
+class Model
+{
+    public HashSet<Dog> ExpectedServerDogs = new HashSet<Dog>();
+    public HashSet<Cat> ExpectedServerCats = new HashSet<Cat>();
+    
+    public HashSet<Dog> ExpectedClientDogs = new HashSet<Dog>();
+    public HashSet<Cat> ExpectedClientCats = new HashSet<Cat>();
```
kazimuth: If you want these in the model, you'll want e.g.:

```csharp
List<Func<Model, HashSet<Dog>>> activeDogSubscriptions = new();
public HashSet<Dog> ExpectedClientDogs {
    get {
        HashSet<Dog> result = new();
        foreach (var activeDogSub in activeDogSubscriptions) {
           result.UnionWith(activeDogSub(this));
        } 
        return result;
     }
}
```

kazimuth:
@rekhoff, I moved your code to the branch `rekhoff/multiplicity-example` and rebased it. Would appreciate if you could open a PR with that code based at this branch.

src/MultiDictionary.cs
joshua-spacetime: Excellent, thanks @kazimuth

gefjon:
John: What's the story on the subscription helper?
James: That can go away now. Also, I want to test this against BitCraft.

rekhoff:
> @rekhoff, I moved your code to the branch `rekhoff/multiplicity-example` and rebased it. Would appreciate if you could open a PR with that code based at this branch.

[PR 254](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/254) opened for multiplicity test.

kazimuth:
Now based on https://github.com/clockworklabs/SpacetimeDB/pull/2303

examples~/quickstart/client/Program.cs
```diff
@@ -123,25 +124,15 @@ void OnConnect(DbConnection conn, Identity identity, string authToken)
...
-        // worry about thread synchronization or anything like that.
-        subscriptions += 1;
-
-        if (subscriptions == 2)
-        {
-            OnSubscriptionApplied(ctx);
-        }
-    };
-
     var userSubscription = conn.SubscriptionBuilder()
```
rekhoff: This now contains both a user and message subscription, to variable maybe wants to be changed from `userSubscription` to be more clear.

tests~/MultiDictionaryTests.cs
```diff
@@ -0,0 +1,239 @@
...
+    [Fact]
+    public void Equality()
+    {
+        ListWithNormalizedDuplicates(Gen.Byte[1, 10], Gen.Byte[1, 10], EqualityComparer<byte>.Default).Sample(list =>
+        {
+            var m1 = MultiDictionary<byte, byte>.FromEnumerable(list, EqualityComparer<byte>.Default, EqualityComparer<byte>.Default);
+            Gen.Shuffle(list);
+            var m2 = MultiDictionary<byte, byte>.FromEnumerable(list, EqualityComparer<byte>.Default, EqualityComparer<byte>.Default);
+
+            Assert.Equal(m1, m2);
```
kazimuth, line 50: The code inside `Sample` is what actually matters in these tests. Basically, `list` is a list of `KeyValuePair<byte, byte>` that is randomly generated, and we test that no matter what value it has, this code works.

tests~/MultiDictionaryTests.cs
```diff
@@ -0,0 +1,239 @@
...
+    /// Generate a list of KeyValuePairs.
+    /// If any two items of the list have duplicate Keys, they are guaranteed to have duplicate Values.
+    /// </summary>
+    /// <typeparam name="TKey"></typeparam>
+    /// <typeparam name="TValue"></typeparam>
+    /// <param name="g1"></param>
+    /// <param name="g2"></param>
+    /// <param name="equality"></param>
+    /// <returns></returns>
+    Gen<List<KeyValuePair<TKey, TValue>>> ListWithNormalizedDuplicates<TKey, TValue>(Gen<TKey> g1, Gen<TValue> g2, IEqualityComparer<TKey> equality, int ListMinLength = 0, int ListMaxLength = 32)
```
kazimuth, line 18: The only thing that really matters about this is that it returns `Gen<List<KeyValuePair<TKey, TValue>>>` which is an object that randomly generates `List<KeyValuePair<TKey, TValue>>`s.

rekhoff (approved):
Reviewed code. Looks correct and well commented. No issues found.

src/MultiDictionary.cs
```diff
@@ -0,0 +1,434 @@
...
+            {
+                var theirDelta = (int)entry.Value.Inserts - (int)entry.Value.Removes;
+                if (theirDelta >= 0)
+                {
+                    return false;
+                }
+                if (self.RawDict.TryGetValue(entry.Key, out var mine))
+                {
+                    var resultMultiplicity = (int)mine.Multiplicity + theirDelta;
+                    return resultMultiplicity <= 0;
```
joshua-spacetime, line 185: Note, it should never be the case that `resultMultiplicity < 0`. I would at least log it if it does happen.

src/MultiDictionary.cs
```diff
@@ -0,0 +1,434 @@
...
+                            RawDict[key] = (their.Value, (uint)newMultiplicity);
+                            wasUpdated.Add((key, my.Value, their.Value)); // store both the old and new values.
+                        }
+                    }
+                    else // if (newMultiplicity <= 0)
+                    {
+                        // This is a removal.
+                        if (newMultiplicity < 0)
+                        {
+                            PseudoThrow($"Internal error: Removing row with key {key} {-theirDelta} times, but it is only present {my.Multiplicity} times.");
```
joshua-spacetime, line 238: Just like you're doing here actually 👍

joshua-spacetime (approved):
I've reviewed the MultiDictionary and Table cache changes for tracking multiplicities and invoking row callbacks. Just a small logging nit, but otherwise the logic is correct.

As for the rest of the changes, I'll defer to @rekhoff.

PR #258, opened 2025-02-27
Update C# server and client quickstart examples to match tutorial document.

rekhoff:
# Description of Changes
Update the C# server and client SDK quickstart-chat example to match the code presented in the tutorial, as of https://github.com/clockworklabs/spacetime-docs/pull/170 .
Also renamed the directory from `quickstart` to `quickstart-chat` in order to be more specific.

# API and ABI breaking changes

If this is an API or ABI breaking change, please apply the
corresponding GitHub label.

# Expected complexity level and risk

1

## Requires SpacetimeDB PRs
com.clockworklabs.spacetimedbsdk: master

# Testing

*Describe any testing you've done, and any testing you'd like your reviewers to do,
so that you're confident that all the changes work as expected!*

- [x] Ran `quickstart-chat` C# server and C# client locally and:
  - Set my name.
  - Sent a message.
  - Restarted and viewed the message backlog.
  - Sent a few more messages.
  - Exited, deleted my local token, restarted and connected as a new identity.
  - Set my new identity's name.
  - Sent a message as my new identity.


tests~/SnapshotTests.cs
```diff
@@ -261,7 +261,7 @@ private static TableUpdate SampleUserUpdate(string identity, string? oldName, st
     private static Message SampleMessage(string identity, long sentMicrosecondsSinceUnixEpoch, string text) => new()
     {
         Sender = Identity.From(Convert.FromBase64String(identity)),
-        Sent = new Timestamp(sentMicrosecondsSinceUnixEpoch),
```
kazimuth: Is this an automatic conversion to Timestamp?

rekhoff: Nope, this change was actually just reverted. The "Sent" parameter of Message was mistakenly a `long`, rather than `Timestamp`.

examples~/quickstart-chat/client/Program.cs
```diff
@@ -5,54 +5,101 @@
...
 void Main()
 {
+    // Initialize the `AuthToken` module
     AuthToken.Init(".spacetime_csharp_quickstart");
-
-    // TODO: just do `var conn = DbConnection...` when OnConnect signature is fixed.
+    // Builds and connects to the database
     DbConnection? conn = null;
+    conn = ConnectToDB();
+    // Registers to run in response to database events.
```
kazimuth:
```suggestion
    // Registers callbacks to run in response to database events.
```

examples~/quickstart-chat/client/Program.cs
```diff
@@ -5,54 +5,101 @@
...
+    RegisterCallbacks(conn);
+    // Declare a threadsafe cancel token to cancel the process loop
```
kazimuth, line 24: It would be nice to have a comment at the top of the file documenting which threads are doing what, and describing the general architecture. There's just two threads.

(The quickstart-chat document currently just says "we'll explain what these are for later", but I think it would be better to start by just saying "we're going to have two threads, one for processing database updates and one for handling user inputs." Maybe this has already been litigated elsewhere though.)

examples~/quickstart-chat/client/Program.cs
```diff
@@ -5,54 +5,101 @@
...
+DbConnection ConnectToDB()
+{
+    DbConnection? conn = null;
     conn = DbConnection.Builder()
         .WithUri(HOST)
         .WithModuleName(DBNAME)
-        //.WithToken(AuthToken.Token)
-        .OnConnect(OnConnect)
+        .WithToken(AuthToken.Token)
+        .OnConnect(OnConnected)
```
kazimuth, line 50: If you're doing this change, shouldn't OnDisconnect be called OnDisconnected?

examples~/quickstart-chat/client/Program.cs
```diff
@@ -5,54 +5,101 @@
 using SpacetimeDB;
 using SpacetimeDB.Types;
 
-const string HOST = "http://localhost:3000";
-const string DBNAME = "chatqs";
 
 // our local client SpacetimeDB identity
 Identity? local_identity = null;
```
kazimuth, line 10: These should ideally be camelCased I think? Maybe they aren't in the tutorial, but if so I'm not sure why?

examples~/quickstart-chat/client/Program.cs
```diff
@@ -80,6 +129,15 @@ void User_OnUpdate(EventContext ctx, User oldValue, User newValue)
     }
 }
 
+/// Our `Message.OnInsert` callback: print new messages.
+void Message_OnInsert(EventContext ctx, Message insertedValue)
+{
+    if (ctx.Event is not Event<Reducer>.SubscribeApplied)
```
kazimuth, line 138: Comment documenting why we shouldn't print them in this case

kazimuth (approved):
Looks good, all the added comments are great. A few small nits.

examples~/quickstart-chat/client/Program.cs
rekhoff: This was likely done because in C#, constants are typically named using all capital letters, with words separated by underscores to improve readability.

examples~/quickstart-chat/client/Program.cs
kazimuth: I think we are not printing them because we don't want to re-print messages from history.

rekhoff: PR updated with a code comment to reflect that.

examples~/quickstart-chat/client/Program.cs
rekhoff: `OnDisconnect` updated to `OnDisconnected`

examples~/quickstart-chat/client/Program.cs
rekhoff: Added a bit more context, replacing "we'll explain what these are for later" with more information as to what we are doing there, and where we will be drilling in deeper.

PR #264, opened 2025-03-07
Fix for unsubscribe bug + regression tests

### LINKED ISSUE #279: C# client SDK: `Unsubscribe` and `UnsubscribeThen` apparently non-functional
[Per a report on Discord](https://discord.com/channels/1037340874172014652/1353949361746477107/1353949361746477107), `Unsubscribe` is apparently nonfunctional in the C# client SDK. The reporter [went as far as to provide a reproduction](https://github.com/JesseDunlap/Blackholio/commit/8be3acf677ff66c1dc108a2b32a0326797ca766f). If this is indeed broken, it's a serious problem that will have major impacts on BitCraft performance.

Definition of Done:

- We have a test which verifies that, in the C# client SDK, `UnsubscribeThen` callbacks are eventually called, and that unsubscribed rows are deleted from the client cache as expected.

kazimuth:
## Description of Changes

This PR fixes one issue and adds regression tests for it, plus another issue. These regression tests are run on CI. You can also run them locally -- run a local spacetime instance and then `tools~/run-regression-tests.sh ../SpacetimeDB`).

### Issue 1: Unsubscribe is non-functional

See https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/issues/279. This was a coding mistake made during the rush to 1.0. It was easily fixed.

### Issue 2: Indexes not tracking deletes

This was recently discovered by the BitCraft team.

The problem was introduced in https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/commit/20c6480424b180522bd95630258b4e4e3b84b607 which introduced `RemoteTableHandle<EventContext, Row>.IndexBase<Column>`. This class stores a `Dictionary<Column, List<Row>>` to allow fast filtering. It relies on `List<Row>.Remove` to remove elements from a list when a row delete is received. `List<Row>.Remove` relies on `Row.Equals` to determine when elements are equal.

`Row.Equals` formerly was not implemented for `[SpacetimeDB.Type]`s or for generated Row types. So, this didn't work.

Now it does: it was fixed by https://github.com/clockworklabs/SpacetimeDB/pull/239 . So all this PR does to address this issue is add a regression test.

## API

The interface IDbConnection was changed; however, this changed part of the interface is `internal`, so this should be entirely invisible to users.

## Requires SpacetimeDB PRs
https://github.com/clockworklabs/SpacetimeDB/pull/239 (merged)

## Testsuite

## Testing

- [x] Added a repro for both issues
- [x] Added the repro to CI
- [x] Fixed both issues


examples~/btree-repro/client/Program.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        Log.Info("Got Delete callback");
+        waiting--;
+        ValidateBTreeIndexes(ctx);
+    };
+}
+
+const uint MAX_ID = 10;
+
+void ValidateBTreeIndexes(IRemoteDbContext conn)
```
joshua-spacetime: Can you add a comment somewhere in this file explaining the purpose of the repro? Or include a link to the github issue?

joshua-spacetime (approved):

joshua-spacetime:
Is this repro something that can be easily turned into an automated regression test? If not, I think we should at least create a follow up ticket to do it.

kazimuth:
This is not currently an automated regression test. It could be made into one without too much work I think. That needs to happen before this merges.

joshua-spacetime (approved):
LGTM

tools~/run-regression-tests.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# This script requires a running local SpacetimeDB instance.
+
+set -ueo pipefail
+
+STDB_PATH="$1"
+SDK_PATH="$(dirname "$0")/.."
+SDK_PATH="$(realpath "$SDK_PATH")"
+
+cargo run --manifest-path "$STDB_PATH/crates/cli/Cargo.toml" -- generate -y -l csharp -o "$SDK_PATH/examples~/regression-tests/client/module_bindings" --project-path "$SDK_PATH/examples~/regression-tests/server"
```
bfops: should this be calling `gen-regression-tests.sh`?

kazimuth: I had been not doing that for some reason but now I think it's fine

PR #269, opened 2025-03-11
Update DLLs to 1.0.1

bfops:
## Description of Changes
I don't know if the DLLs need updating, but updated the DLLs just in case.

## API

Incorporate any chances in SpacetimeDB 1.0.1. There should be no breaking changes.

## Requires SpacetimeDB PRs

## Testsuite
SpacetimeDB branch name: release/v1.0.1

## Testing
None. Will test in release.


jdetter (approved):
Thanks Zeke 👍

PR #271, opened 2025-03-13
Adds tests of GenericEqualityComparer

rekhoff:
Adds an additional test of the `GenericEqualityComparer` function for `Integer`, `String`, `Byte Array`, `Identity`, and `User` type comparisons.

This PR resolves additional checks as reported in by [Issue 230](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/issues/230) that where not-yet covered elsewhere.

## API

## Requires SpacetimeDB PRs
No other PR are needed for this to work.

## Testsuite

## Testing
- [X] Locally ran `dotnet test` against the `tests.csproj` project and it passed all tests

### DIFF
--- a/tests~/Tests.cs
+++ b/tests~/Tests.cs
@@ -7,7 +7,7 @@ public class Tests
     [Fact]
     public static void GenericEqualityComparerCheck()
     {
-        // Tests the behavior of the GenericEqualityComparer function
+        // Validates the behavior of the GenericEqualityComparer's Equals function
 
         // Byte Arrays
         byte[] byteArray = new byte[10];
@@ -21,12 +21,10 @@ public class Tests
 
         // Integers
         int integer = 5;
-        int integerByRef = integer;
         int integerByValue = 5;
         int integerUnequalValue = 7;
         string integerAsDifferingType = "5";
 
-        Assert.True(GenericEqualityComparer.Instance.Equals(integer, integerByRef));
         Assert.True(GenericEqualityComparer.Instance.Equals(integer, integerByValue));
         Assert.False(GenericEqualityComparer.Instance.Equals(integer, integerUnequalValue));
         // GenericEqualityComparer does not support to converting datatypes and will fail this test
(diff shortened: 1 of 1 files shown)


tests~/Tests.cs
```diff
@@ -0,0 +1,68 @@
...
+        byte[] byteArrayByValue = new byte[10];
+        byte[] byteArrayUnequalValue = new byte[01];
+
+        Assert.True(GenericEqualityComparer.Instance.Equals(byteArray, byteArrayByRef));
+        Assert.True(GenericEqualityComparer.Instance.Equals(byteArray, byteArrayByValue));
+        Assert.False(GenericEqualityComparer.Instance.Equals(byteArray, byteArrayUnequalValue));
+
+        // Integers
+        int integer = 5;
+        int integerByRef = integer;
```
kazimuth: Ints are value types so this is a value copy, not a reference copy. So this does exactly the same thing as the next line.

tests~/Tests.cs
```diff
@@ -0,0 +1,68 @@
...
+        Assert.True(GenericEqualityComparer.Instance.Equals(identity, identityByRef));
+        Assert.True(GenericEqualityComparer.Instance.Equals(identity, identityByValue));
+        Assert.False(GenericEqualityComparer.Instance.Equals(identity, identityUnequalValue));
+
+        Assert.True(GenericEqualityComparer.Instance.Equals(testUser, testUserByRef));
+        Assert.True(GenericEqualityComparer.Instance.Equals(testUser, testUserByValue));
+        Assert.False(GenericEqualityComparer.Instance.Equals(testUser, testUserUnequalIdentityValue));
+        Assert.False(GenericEqualityComparer.Instance.Equals(testUser, testUserUnequalNameValue));
+        Assert.False(GenericEqualityComparer.Instance.Equals(testUser, testUserUnequalOnlineValue));
+    }
```
kazimuth, line 83: Would also like to check that it compares `TaggedEnum` values correctly.

tests~/Tests.cs
rekhoff: Ah geez, right. Removed.

tests~/Tests.cs
rekhoff: `TaggedEnum` is an abstract type. Added a `Status` record test, because `Status` inherits from `TaggedEnum`, thus providing a valid test of `TaggedEnum`.

kazimuth (approved):
👍

PR #285, opened 2025-04-03
Regression tests updated to point to Runtime version 1.0.1

jdetter:
## Description of Changes

## API

## Requires SpacetimeDB PRs

## Testsuite

## Testing

### DIFF
--- a/examples~/regression-tests/server/StdbModule.csproj
+++ b/examples~/regression-tests/server/StdbModule.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="SpacetimeDB.Runtime" Version="1.1.0" />
+    <PackageReference Include="SpacetimeDB.Runtime" Version="1.0.1" />
   </ItemGroup>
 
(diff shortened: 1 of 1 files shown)


examples~/regression-tests/server/StdbModule.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="SpacetimeDB.Runtime" Version="1.1.0" />
```
jdetter, line 11: This version doesn't exist on nuget so running `dotnet build` locally doesn't work.

bfops: wait, wasn't the original bugfix inside the DLLs? Shouldn't we instead be pointing to the latest version of the SpacetimeDB repo using a `nuget.config` so that we're actually regression-testing the current thing? (Otherwise we'd be running regression tests against a "frozen" version of SpacetimeDB, no?)

bfops:
closing this since 1.1.0 is now released

PR #286, opened 2025-04-06
SpacetimeDB working in Unity WebGL Builds

### LINKED ISSUE #278: Support for Unity WebGL builds
Currently WebGL builds are not supported because SpacetimeDB uses threads to process incoming network messages. I think there are several solutions here:

1. We could use `Task` everywhere instead of threads - we have to determine the performance impact of doing this. It needs to be tested against BitCraft.
2. We could use `#ifdef` to only use threads for non-webgl platforms and then use `Task` instead when we are building for WebGL.

First let's just try #1 and then if that doesn't work well then let's go with #2.

Steps to move this forward
- [ ] Try getting this working with a webgl build in blackholio https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/274
- [ ] ^ If that works, then try getting it working with BitCraft (not in webgl, just standalone) - then determine the performance impact of this.

Daxode:
## Description of Changes
As it stands today, Unity WebGL doesn't work. Partially the reason for this is Multi-Threading, and the other reason is the use of `ClientWebSocket`. In order to fix this (specifically in the case of Unity), here's some changes that _can_ be made. Note that this is mostly a suggestion and does come with a few flaws, though arguably it might still be better than it not working at all? Up to you!

The Tl;Dr of how it works, is to:
- **MultiThreading Problem**: simply invoke the `Task.Run` functions on main thread instead, and use a coroutine in place of where the two simultaneous threads was expected.
- **ClientWebSocket**: Use a `.jslib` to create the WebSocket directly within Javascript, and then have JS call the corresponding correct functions.

DISCLAIMER: currently OnClose doesn't quite work correctly as `__allocate` isn't invoked correctly. 

## API

Not a breaking change to the API, should be internal implementation details

## Requires SpacetimeDB PRs
None

## Testsuite
????

## Testing
Open the Blackholeio project, try building it for WebGL

- [X] Made a game using the feature: https://daxode.itch.io/eat-to-the-deep


bfops:
Thanks for submitting this! We'll try to get this reviewed and get back to you.

jdetter (approved):
I have tested + reviewed these changes, I've fixed the disconnect issue and I don't have any remaining issues to block this from merging.

I would like @kazimuth to take a look and do a code review before this merges.

Thanks again @Daxode! 😄

rekhoff (approved):
Reviewed changes. Everything checks out. Tested and validated functionality. Thanks @Daxode!

kazimuth:
Looks like we need to run `dotnet format SpacetimeDB.ClientSDK.sln` to get the tests passing, I can do that tomorrow morning.

Reading through this it looks sensible. It would be nice to be able to automatically test this behavior but unfortunately that is going to require a fairly complicated harness, so I'm okay merging without one. I'm going to do a closer read tomorrow morning but I think I'm happy with this. Thanks so much @Daxode !

src/Plugins/WebSocket.jslib
```diff
@@ -0,0 +1,98 @@
...
+
+            socket.onmessage = function(event) {
+                if (manager.callbacks.message && event.data instanceof ArrayBuffer) {
+                    var buffer = _malloc(event.data.byteLength);
+                    HEAPU8.set(new Uint8Array(event.data), buffer);
+                    dynCall('viii', manager.callbacks.message, [socketId, buffer, event.data.byteLength]);
+                    _free(buffer);
+                }
+            };
+            var __allocate = this.allocate;
```
kazimuth: This is unused -- is this related to what you said about `allocate` not being invoked correctly?

kazimuth (commented):
Overall this looks good, the bindings glue makes sense. I'm curious how we can fix the `OnClose` issue you mentioned.

src/Plugins/WebSocket.jslib
Daxode: @jdetter fixed the close issue, but forgot to remove this part, it can be removed safely, and the allocate issue he already fixed by correctly using malloc

kazimuth (approved):
👍

Daxode:
Hmm.. seems test fail due to Unity Licensing? How should we proceed? 🤔

bfops:
Thanks for flagging that! We'll look into it.

PR #294, opened 2025-04-16
Fix row deduplication issues

kazimuth:
## Description of Changes
Addresses https://github.com/clockworklabs/SpacetimeDBPrivate/issues/1585

There was an issue with row deduplication when a TransactionUpdate of the form:

```
+ k1: v1
- k1: v1
- k1: v1
+ k1: v2
```
appeared, where `k1` is the primary key of a row and `v1`, `v2` are the full values of that row. The old manner of processing updates would see this as a no-op. I've updated the logic to handle this case correctly, at the cost of an extra equality comparison per row in an update. (Actually it's slightly cheaper than this, it skips the equality comparison if there is only row with a particular primary key in an update. This may help speed up processing of large updates e.g. chunk data in BitCraft.)

## API

## Testsuite

## Testing

- [x] Added a failing test, then fixed it

### DIFF
--- a/src/MultiDictionary.cs
+++ b/src/MultiDictionary.cs
@@ -74,7 +74,7 @@ namespace SpacetimeDB
             Debug.Assert(key != null);
             if (RawDict.TryGetValue(key, out var result))
             {
-                Debug.Assert(ValueComparer.Equals(value, result.Value), "Added key-value pair with mismatched value to existing data");
+                Debug.Assert(ValueComparer.Equals(value, result.Value), $"Added key-value pair with mismatched value to existing data, {key} {value} {result.Value}");
                 RawDict[key] = (value, result.Multiplicity + 1);
                 return false;
             }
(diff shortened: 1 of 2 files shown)


src/MultiDictionary.cs
```diff
@@ -316,19 +347,237 @@ public override string ToString()
...
+        ///     get the values before and after the change, along with their multiplicities.
+        ///     
+        /// - If the value associated to the key does not change, the key-value pair can still have multiplicity changes.
+        ///     Use <c>.NonValueChange</c> to get at this multiplicity information.
+        /// 
+        /// </summary>
+        public struct KeyDelta
+        {
+            // The (one | two) values associated with this key.
+            // Invariant: If D2 is present, its Delta should be <= the Delta for D1.
```
joshua-spacetime: Can you expand on this? Is this because `D1` represents the old value and so has a multiplicity < 0, whereas `D2` represents the new value and so has a multiplicity > 0? If so, we should make it clear that we're talking about the signed multiplicities, but not necessarily their magnitudes.

tests~/MultiDictionaryTests.cs
```diff
@@ -241,4 +241,25 @@ public void IdentitiesWorkAsPrimaryKeys()
...
+    public void InsertThenDeleteOfOldRow()
+    {
+        var dict = new MultiDictionary<byte, byte>(EqualityComparer<byte>.Default, EqualityComparer<byte>.Default);
+        dict.Add(1, 2);
+
+        var delta = new MultiDictionaryDelta<byte, byte>(EqualityComparer<byte>.Default, EqualityComparer<byte>.Default);
+        delta.Add(1, 2);
+        delta.Add(1, 3);
+        delta.Remove(1, 2);
+        delta.Remove(1, 2);
```
joshua-spacetime, line 378: I think it would be useful to test different permutations of these operations. I also think we should add a test for inserts and deletes of both the old and new value.

src/MultiDictionary.cs
kazimuth: Wrote a longer comment

tests~/MultiDictionaryTests.cs
kazimuth: Added a proptest that tests large chunks of operations with value changes. I also added one more permutation to this test.

joshua-spacetime (approved):

PR #300, opened 2025-04-22
Update DLLs to 1.1.1

bfops:
## Description of Changes
Just bumping the DLLs for a version number update. No actual changes.

## API

No breaking changes

## Testsuite

## Testing
Just existing CI. This only bumps versions.


jdetter (approved):
I was going to ask to have Ryan test this but since this runs Unity in CI I think this should actually be fine. Thanks Zeke 👍 

Also, we deleted the previous versions of the DLLs correct?

bfops:
> I was going to ask to have Ryan test this but since this runs Unity in CI I think this should actually be fine. Thanks Zeke 👍
> 
> Also, we deleted the previous versions of the DLLs correct?

Correct (that's why they show as renamed rather than added).

jdetter:
> > I was going to ask to have Ryan test this but since this runs Unity in CI I think this should actually be fine. Thanks Zeke 👍
> > Also, we deleted the previous versions of the DLLs correct?
> 
> Correct (that's why they show as renamed rather than added).

I see, thanks 👍

bfops:
huh, I'm not sure why it thinks those meta files are wrong..

bfops:
well, it didn't trigger again on the latest run so :shrug:

PR #302, opened 2025-04-23
Pre-decompress query updates

kazimuth:
This dramatically improves performance by avoiding the default implementation of BrotliStream.ReadByte() inherited from Stream, which allocates an array per byte read.

## API

## Testsuite

## Testing


kazimuth:
I expect tests to fail here until https://github.com/clockworklabs/SpacetimeDB/pull/2636 is finished merging.

rekhoff (approved):
Perfect. This was already validated during a live coding session, and the changes in the PR reflect that change. Looks great, approved!

PR #305, opened 2025-04-24
Significantly reduce small byte array allocations

kazimuth:
This purges the DbValue type, instead using row instances themselves as primary key for rows without primary keys. In addition, it instantiates only a single BinaryReader when reading updates for a table, rather than instantiating a BinaryReader and performing an array copy per-row of the table.

Addresses https://github.com/clockworklabs/SpacetimeDBPrivate/issues/1633

## API

## Testsuite

## Testing

- [ ] CI

### DIFF
--- a/src/SpacetimeDBClient.cs
+++ b/src/SpacetimeDBClient.cs
@@ -211,11 +211,10 @@ namespace SpacetimeDB
 
         struct ProcessedDatabaseUpdate
         {
-            // Map: table handles -> (primary key -> DbValue).
-            // If a particular table has no primary key, the "primary key" is just a byte[]
-            // storing the BSATN encoding of the row.
-            // See Decode(...).
-            public Dictionary<IRemoteTableHandle, MultiDictionaryDelta<object, DbValue>> Updates;
+            // Map: table handles -> (primary key -> IStructuralReadWrite).
+            // If a particular table has no primary key, the "primary key" is just the row itself.
+            // This is valid because any [SpacetimeDB.Type] automatically has a correct Equals and HashSet implementation.
+            public Dictionary<IRemoteTableHandle, MultiDictionaryDelta<object, IStructuralReadWrite>> Updates;
 
             // Can't override the default constructor. Make sure you use this one!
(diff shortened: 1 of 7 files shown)


src/Table.cs
```diff
@@ -138,10 +140,10 @@ public RemoteTableHandle(IDbConnection conn) : base(conn) { }
         // We store the BSATN encodings of objects next to their runtime representation.
         // This is memory-inefficient, but allows us to quickly compare objects when seeing if an update is a "real"
         // update or just a multiplicity change.
-        private readonly MultiDictionary<object, DbValue> Entries = new(GenericEqualityComparer.Instance, DbValueComparer.Instance);
+        private readonly MultiDictionary<object, IStructuralReadWrite> Entries = new(GenericEqualityComparer.Instance, GenericEqualityComparer.Instance);
 
         // The function to use for decoding a type value.
-        IStructuralReadWrite IRemoteTableHandle.DecodeValue(byte[] bytes) => BSATNHelpers.Decode<Row>(bytes);
+        IStructuralReadWrite IRemoteTableHandle.DecodeValue(BinaryReader reader) => IStructuralReadWrite.Read<Row>(reader);
```
joshua-spacetime, line 146: Can we remove `BSATNHelpers.Decode` by any chance? Is it still used elsewhere? If so, can you add a comment that it shouldn't be called on hot paths?

joshua-spacetime (approved):

src/SpacetimeDBClient.cs
```diff
@@ -268,15 +267,17 @@ struct ProcessedMessage
         /// <param name="bin"></param>
```
rekhoff:
```suggestion
        /// <param name="reader"></param>
```

rekhoff (approved):

src/Table.cs
kazimuth: It's in SpacetimeDB unfortunately, but I can do that in a separate PR.

kazimuth: Wait, no it isn't. It's used in some places that make sense though. I added a comment.

PR #306, opened 2025-04-25
Convert problem Linq calls to custom enumerators

jdetter:
## Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDBPrivate/issues/1647.

This fixes a performance issue where iterating through the BSATN bytes of a subscribe call was very slow. The problematic function was `BsatnRowListIter` in particular.
```cs
private static IEnumerable<byte[]> BsatnRowListIter(BsatnRowList list)
{
    var rowsData = list.RowsData;

    return list.SizeHint switch
    {
        RowSizeHint.FixedSize(var size) => Enumerable
            .Range(0, rowsData.Count / size)
            .Select(index => rowsData.Skip(index * size).Take(size).ToArray()),

        RowSizeHint.RowOffsets(var offsets) => offsets.Zip(
            offsets.Skip(1).Append((ulong)rowsData.Count),
            (start, end) => rowsData.Take((int)end).Skip((int)start).ToArray()
        ),

        _ => throw new InvalidOperationException("Unknown RowSizeHint variant"),
    };
}
```

And within this function, the problem was specifically the Linq calls `Skip` and `Take`.

Note that `Skip` by default is a linear time operation - reading `rowsData` from the beginning to the passed index. In this case, that means reading the bytes of an entire table from the beginning just to read the next row.

Mono's jit compiler is clever enough to recognize the `Skip().Take()` combination and optimize this completely way so that it only reads the bytes of each row on each call, resulting in a linear scan of the table.

IL2CPP's aot compiler however cannot do this, and so it resorts to the default algorithm which results in a quadratic time scan of the table.

Ultimately the fix was to not rely on the runtime/compiler to optimize declaratively defined Linq code, but rather to define an imperative linear-time algorithm.

## API

Not API breaking.

## Requires SpacetimeDB PRs

- None

## Testsuite

## Testing

- [x] Tested against BitCraft and the performance regressions are gone.

### DIFF
--- a/src/SpacetimeDBClient.cs
+++ b/src/SpacetimeDBClient.cs
@@ -391,20 +391,20 @@ namespace SpacetimeDB
             {
                 if (_index >= _maxIndex)
                     return false;
-                    
+
                 _index++;
-                
+
                 // Create the array only when requested
                 var startIndex = _index * _size;
                 if (startIndex + _size > _rowsData.Count)
                     return false;
-                    
+
                 _current = new byte[_size];
                 for (var i = 0; i < _size; i++)
                 {
                     _current[i] = _rowsData[startIndex + i];
                 }
-                
+
                 return true;
             }
 
@@ -442,23 +442,23 @@ namespace SpacetimeDB
                 if (_index >= _offsets.Count - 1)
                     return false;
                 _index++;
-                
+
                 // Determine start and end indices
                 var startIndex = (int)_offsets[_index];
-                var endIndex = (_index + 1 < _offsets.Count) 
-                    ? (int)_offsets[_index + 1] 
+                var endIndex = (_index + 1 < _offsets.Count)
+                    ? (int)_offsets[_index + 1]
                     : _rowsData.Count;
-                    
+
                 var length = endIndex - startIndex;
                 if (length <= 0)
                     return false;
-                    
+
                 _current = new byte[length];
                 for (var i = 0; i < length; i++)
                 {
                     _current[i] = _rowsData[startIndex + i];
                 }
-                
+
                 return true;
             }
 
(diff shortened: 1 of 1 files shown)


cloutiertyler:
@jdetter could you please update the PR description with more context on the performance issue, it's impact, and how we found it. Also could you please update the fix to include a comment in the code with the context on the issue, so that we don't have a regression in the future?

Alternatively, @joshua-spacetime could you do this?

src/SpacetimeDBClient.cs
```diff
@@ -357,21 +357,125 @@ private static IEnumerable<byte[]> BsatnRowListIter(BsatnRowList list)
         {
             var rowsData = list.RowsData;
 
+            // Previously here we were using LINQ to do what we're now doing with the custom row iterators. This
+            // is due to the fact that the LINQ statements were fast under the Mono runtime, but *a lot* slower
+            // under IL2CPP. This has to do with differences between how Mono uses JIT VS how IL2CPP uses AOT.
+            // Apparently Mono's JIT is smart enough to inline most of the LINQ ops but IL2CPP's AOT is not.
```
joshua-spacetime, line 363:
```suggestion
            // Apparently Mono's JIT is smart enough to inline most of the LINQ ops but IL2CPP's AOT is not.
            // In this case specifically, it was that Mono could optimize `Skip().Take()`,
            // resulting in a linear scan of `rowsData`, whereas IL2CPP could not, resulting in a quadratic scan.
```

joshua-spacetime:
https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/305 fixed this issue indirectly by also removing the Linq calls, so we can close this one. Although the comment would still be valuable.

joshua-spacetime:
Because #305 addresses, and fixed, the same performance issue as this patch, I'm closing this one in favor of #309 which adds the comment.

PR #315, opened 2025-05-12
Faster deserialization

kazimuth:
## Description of Changes
Leverages https://github.com/clockworklabs/SpacetimeDB/pull/2725 to speed up row deserialization. Also updates DLLs for recent upstream fixes.

Changes:
- Uses much simpler code when deserializing enums, avoiding the use of reflection
- Uses manually monomorphized row deserialization methods, which again avoid the use of reflection when constructing row objects

## API

## Requires SpacetimeDB PRs
https://github.com/clockworklabs/SpacetimeDB/pull/2725

## Testsuite

SpacetimeDB branch name: jgilles/perf/faster-allocation

## Testing

- [x] Blackholio
- [x] Bitcraft (have done a basic test but no in-depth testing)


kazimuth:
Test failures seem to be because Unity is getting more aggressive with their license enforcement...

kazimuth:
I have tested this in BitCraft and am happy with it.

src/Table.cs
```diff
@@ -134,16 +146,35 @@ public RemoteTableHandle(IDbConnection conn) : base(conn) { }
         // THE DATA IN THE TABLE.
         // The keys of this map are:
         // - Primary keys, if we have them.
-        // - Byte arrays, if we don't.
+        // - The entire row itself, if we don't.
```
kazimuth, line 149: This has been true for about a month, I forgot to update the comment.

rekhoff (approved):
The reasoning here makes sense and is a well documented justification. I don't seen any issues with this, approved.

PR #316, opened 2025-05-12
Fix list equals

### LINKED ISSUE #313: Major bug: equality comparison for rows is broken in latest master
Current master should not be released until this is fixed.

The specific issue is that BSATN.Codegen generates incorrect `Equals` implementations for types containing lists. This did not matter until https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/commit/a75a5520adc36f5d90d1c26c9bbb9e2dc7b862e1, since before that PR whole-row `.Equals` was not used; rows were instead compared by the equality on their serialization to byte arrays. (E: except in indices, which caused BitCraft problems.) This needs to be fixed in `SpacetimeDB` with a change to BSATN.Codegen, followed by a PR here to update the DLLs.

kazimuth:
## Description of Changes

Update DLLs and generated code for SpacetimeDB 74ad4a14c

## API

## Requires SpacetimeDB PRs

## Testsuite

## Testing


joshua-spacetime (approved):
Approving as this has been tested by BitCraft and was successful.

PR #318, opened 2025-05-16
Pre-hash rows and add SmallHashSets

kazimuth:
## Description of Changes
Annotate rows with their hashes. This is halfway to the old solution of annotating code with their serialized forms; it speeds up insertion into `BTreeIndex`es while saving a byte array allocation per-row.
(Wait, why do we hash when inserting into `BTreeIndex`es, you ask? It's because we use a `HashSet` to store the multiple rows corresponding to a non-unique key.)

This saves work on the main thread repeatedly hashing values. Once we parallelize message processing, this work will be spread out over multiple threads too.

Also adds a data type `SmallHashSet<T>` for use in BTreeIndexes. This is a struct that can store at most one element without allocating. This gives dramatic allocation reduction on initial connection -- it seems that most rows in BTreeIndexes don't have any other rows with the same key. 
Then I manually specialized this, since it's only used on one type -- `PreHashedRow`. using `SmallHashSetOfPreHashedRow` instead of `SmallHashSet<PreHashedRow>` allows us to inline some methods and pass around fewer pointers while hashing things, which is probably good.

## API

## Testsuite

## Testing

I will test:

- [ ] blackholio
- [ ] bitcraft


kazimuth:
This does mildly pessimize calls to Filter -- it adds a second allocation to them.
SmallHashSet is so good I'm tempted to just use it and get rid of PreHashedRow, not sure.

kazimuth:
I've added pooling of HashSets within the SmallHashSet class.

src/SpacetimeDBClient.cs
```diff
@@ -226,7 +226,7 @@ internal struct ParsedDatabaseUpdate
             // Map: table handles -> (primary key -> IStructuralReadWrite).
```
rekhoff, line 226: This comment could use an update, as it still references `IStructuralReadWrite`.

src/SmallHashSet.cs
```diff
@@ -0,0 +1,352 @@
...
+        }
+        else
+        {
+            return new(2, DefaultEqualityComparer);
+        }
+    }
+}
+
+/// <summary>
+/// This is a silly object.
```
rekhoff, line 145: Though the struct is indeed silly, we may want to provide more context or remove this summary.

src/SmallHashSet.cs
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    public void Reset()
+    {
+        state = State.Started;
+    }
+}
+
+/// <summary>
+/// This is a very silly object.
```
rekhoff, line 197: Removal of summary or adding additional context is also needed for this very silly struct.

rekhoff (approved):
Assuming all testing goes well, and there are not any significant performance impact, the changes look good and are in-line with the stated goal of the pull request. I've added some minor notes to some comment lines, but that is all. I'm marking this approved.

kazimuth:
I'm not happy with this yet. I'm not entirely convinced it improves performance; in my testing, allocating HashSets from the pool is taking 5% of time to apply an update. But that's under deep profiling, so I have no idea if it's true or not. We could really use better benchmarks.

bfops:
@kazimuth is doing some more testing on this and then we may merge + release it

bfops:
Migrated to https://github.com/clockworklabs/SpacetimeDB/pull/3146, as we are merging this repo into that one.

PR #322, opened 2025-05-22
Update DLLs to 1.1.2

bfops:
## Description of Changes
Update DLLs to the 1.1.2 versions that have been pushed to NuGet.

The SpacetimeDB tag that they were built from is `v1.1.2-csharp-release`.

## API

## Requires SpacetimeDB PRs
None, because we're updating the DLLs to essentially include SpacetimeDB PRs.

## Testsuite

## Testing
CI I guess?


kazimuth (approved):
👍

PR #326, opened 2025-06-04
Faster Stats.cs

kazimuth:
## Description of Changes
`NetworkRequestTracker` previously was keeping all historical reducer request data, and searching through this every frame to get statistics. I've modified it to throw out much more data -- it's much faster now, but only updates every few seconds.

## API

Adds new APIs.

## Requires SpacetimeDB PRs
N/A

## Testsuite

## Testing

- [x] Tested Bitcraft. **Their F9 debug menu will require an update, since we now only keep one time window of request data, rather than being able to give information about multiple windows.** But it works.
- [x] Blackholio CI


kazimuth:
Fixes https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/issues/323

rekhoff (approved):
Tested machine against BitCraft on the Windows Server machine running in editor with 1024CCU. After this fix, there is still a delay when opening the F9 menu, but once opened, there is no apparent lag from the menu being open.
This indicates to me that the values returned by the F9 menu should be far more credible with this fix. I'm marking this as approved.

kazimuth:
@joshua-spacetime suggested we just discard the previous max every 5 seconds. I'm inclined to do that since it reduces allocator pressure. Gonna rewrite this.

kazimuth:
Now this is rewritten to support multiple windows. Values take longer to show up for the first time, and only update every n seconds, but I think that's fine. This now uses much less time and memory. I've tested against BitCraft again and am ready to merge.

PR #330, opened 2025-06-12
More granular request process tracking

### LINKED ISSUE #329: Add size of message queue to F9 metrics
Similar to SpacetimeDB's reducer queue size metric, we should sample the size of the client's message queue every 5 seconds and add this to the F9 metrics. A point in time sample or a max is sufficient. Tracking the min for this metric is not very useful.

kazimuth:
## Description of Changes
Addresses https://github.com/clockworklabs/SpacetimeDBPrivate/issues/1786 and https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/issues/329. Marks messages as parsed from the background parsing thread, and then marks them as applied using a separate tracker once they are applied.

## API

## Testsuite

## Testing

- [x] Testing against Bitcraft

### DIFF
--- a/src/Stats.cs
+++ b/src/Stats.cs
@@ -137,47 +137,51 @@ namespace SpacetimeDB
             }
         }
 
-        // The remaining methods in this class do not need to lock, since they are only called from OnProcessMessageComplete.
-
         internal bool FinishTrackingRequest(uint requestId)
         {
-            if (!_requests.Remove(requestId, out var entry))
+            lock (this)
             {
-                // TODO: When we implement requestId json support for SpacetimeDB this shouldn't happen anymore!
-                // var minKey = _requests.Keys.Min();
-                // entry = _requests[minKey];
-                //
-                // if (!_requests.Remove(minKey))
-                // {
-                //     return false;
-                // }
-                return false;
-            }
+                if (!_requests.Remove(requestId, out var entry))
+                {
+                    // TODO: When we implement requestId json support for SpacetimeDB this shouldn't happen anymore!
+                    // var minKey = _requests.Keys.Min();
+                    // entry = _requests[minKey];
+                    //
+                    // if (!_requests.Remove(minKey))
+                    // {
+                    //     return false;
+                    // }
+                    return false;
+                }
 
-            // Calculate the duration and add it to the queue
-            InsertRequest(entry.Start, entry.Metadata);
-            return true;
+                // Calculate the duration and add it to the queue
+                InsertRequest(entry.Start, entry.Metadata);
+                return true;
+            }
         }
 
         internal void InsertRequest(TimeSpan duration, string metadata)
         {
-            var sample = (duration, metadata);
+            lock (this)
+            {
+                var sample = (duration, metadata);
 
-            if (AllTimeMin == null || AllTimeMin.Value.Duration > duration)
-            {
-                AllTimeMin = sample;
-            }
-            if (AllTimeMax == null || AllTimeMax.Value.Duration < duration)
-            {
-                AllTimeMax = sample;
-            }
-            _totalSamples += 1;
+                if (AllTimeMin == null || AllTimeMin.Value.Duration > duration)
+                {
+                    AllTimeMin = sample;
+                }
+                if (AllTimeMax == null || AllTimeMax.Value.Duration < duration)
+                {
+                    AllTimeMax = sample;
+                }
+                _totalSamples += 1;
 
-            foreach (var window in TrackerWindows)
-            {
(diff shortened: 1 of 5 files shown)


kazimuth:
I've tested this against BitCraft locally, and opened https://github.com/clockworklabs/BitCraft/pull/5506 which can be merged after this PR is merged and released.

kazimuth:
Screenshot 2025-06-12 165946

It looks like messages wait in the queue to be applied a lot longer than they wait in the queue to be parsed, in the un-loaded case.

kazimuth:
Note that this changes "Reducer Latency" from measuring `call -> apply` time to measuring `call -> receive` time. The order of events is

```
call    ->      receive      ->      parse     ->    apply
     network               client            client
```

src/Stats.cs
```diff
@@ -229,10 +235,41 @@ internal void InsertRequest(DateTime start, string metadata)
 
     public class Stats
     {
+        /// <summary>
+        /// Tracks times from reducers requests being sent to their responses being received.
+        /// This is clientside time, not host-time: for that you want AllReducersTracker.
```
joshua-spacetime: This is `host-time + network + outer parse-time`?

src/Stats.cs
```diff
@@ -229,10 +235,41 @@ internal void InsertRequest(DateTime start, string metadata)
...
+        /// Includes the time needed to parse the outer layer of the response message, but not the inner layer
+        /// if it includes a DatabaseUpdate.
+        /// </summary>
         public readonly NetworkRequestTracker ReducerRequestTracker = new();
+
+        /// <summary>
+        /// Tracks times from one-off requests being sent to their responses being received.
```
joshua-spacetime, line 270: This does not include message parsing/processing? Does it include any time waiting in queue?

src/Stats.cs
```diff
@@ -229,10 +235,41 @@ internal void InsertRequest(DateTime start, string metadata)
...
+        /// </summary>
         public readonly NetworkRequestTracker OneOffRequestTracker = new();
+
+        /// <summary>
+        /// Tracks times from subscriptions being sent to their responses being received.
+        /// Includes the time needed to parse the outer layer of the response message, but not the inner layer
+        /// if it includes a DatabaseUpdate.
```
joshua-spacetime: Does it include time waiting in queue?

src/Stats.cs
```diff
@@ -229,10 +235,41 @@ internal void InsertRequest(DateTime start, string metadata)
...
+
+        /// <summary>
+        /// Tracks times from messages being received on the wire to their being fully parsed,
+        /// but not applied.
+        /// Includes time waiting in the pre-parsing queue.
+        /// </summary>
         public readonly NetworkRequestTracker ParseMessageTracker = new();
+
+        /// <summary>
+        /// Tracks times from messages being parsed on a background thread to their being applied on the main thread.
```
joshua-spacetime, line 298: Does this one include time waiting in queue?

src/Stats.cs
kazimuth: Yes, I added comments clarifying this.
I could break it down further with measurements of time waiting in queue versus actually being parsed. Do you think that would be helpful?

src/Stats.cs
kazimuth: This includes queue and parsing times because OneOffQuery is somewhat weird.

src/Stats.cs
kazimuth: Yes, clarified comment.

src/Stats.cs
kazimuth: Yes, clarified comment.

kazimuth:
I've split the trackers to be more granular, and updated SpacetimeDBClient's internal names to match the names we're using in `Stats.cs`. This should make the code easier to understand; it's not a user-visible change.

joshua-spacetime (approved):
Approving, but with one request. Can you further clarify the difference between `ParseMessageQueueTracker` and `ParseMessageTracker`? Specifically which one includes parse time and which one does not, assuming both include the queue time.

rekhoff (approved):
These changes look great. The code is well commented and provides greater details about how the metrics are calculated. My only request as a follow-up might be to add on the BitCraft side, some timing descriptions to the F9 menu text, so what timing specifically is being shown comes just after the name of the metric. I'll approve this as well.

kazimuth:
> Approving, but with one request. Can you further clarify the difference between `ParseMessageQueueTracker` and `ParseMessageTracker`? Specifically which one includes parse time and which one does not, assuming both include the queue time.

One of them is only queue time, and one is only parse time.

PR #331, opened 2025-06-19
Revert "Fix row deduplication issues (#294)"

kazimuth:
## Description of Changes

This reverts commit [ba9f3be337f7c35198a24e94a67a448dc6570819](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/294).
This commit complicated the logic inside `MultiDictionary` and `MultiDictionaryDelta`, making them slower to build and apply. This was needed to address a particular kind of `TransactionUpdate`:

```
+ k1: v1
- k1: v1
- k1: v1
+ k1: v2
```

I believe the server has no longer sent this kind of update since:
https://github.com/clockworklabs/SpacetimeDB/pull/2654
But I want to double-check with @joshua-spacetime that that is the case.

This PR requires careful testing to ensure that it doesn't resurrect the issues BitCraft was having with chests (really, with out of date clientside indexes).

If we merge this PR, we are effectively **breaking the SDK** for users who are on any version of SpacetimeDB older than the one with Joshua's fix, in the form of resurrecting a dead bug. However, users on newer versions will see better performance. I'm not sure if that's a tradeoff we're willing to make.

## API

## Requires SpacetimeDB PRs

## Testsuite

## Testing


kazimuth:
On further reflection, once BitCraft fixes their style of chunk subscription, there'll be a lot less equality comparisons in the hot path, because they won't be laying a copy of the visible chunks over the existing visible chunks. The main benefit of this change is removing one equality comparison from the hot path, so I think it's fine to not merge this now, since the change they're doing will be a much better way to improve performance, without the chance of regression.

bfops:
Given the last comment, I'm going to close this PR.

PR #349, opened 2025-07-15
Add some developer docs

kazimuth:
## Description of Changes

Add some developer docs.

## API

## Testsuite

## Testing
N/A


kazimuth:
See also https://github.com/clockworklabs/SpacetimeDB/pull/2938

DEVELOP.md
```diff
@@ -17,3 +14,85 @@ dotnet pack ../SpacetimeDB/crates/bindings-csharp/BSATN.Runtime && ./tools~/writ
...
+
+A client created with this SDK is at root a `DbConnection`. This class lives in the generated code in the file `SpacetimeDBClient.g.cs`. See e.g. [`examples~/quickstart-chat/client/module_bindings/SpacetimeDBClient.g.cs`](./examples~/quickstart-chat/client/module_bindings/SpacetimeDBClient.g.cs).
+(Note that `SpacetimeDBClient` is a vestigial name that should probably be retired at some point...)
+
+`DbConnection` in the generated code inherits from `DbConnectionBase<...>` in the SDK code, which lives in [`src/SpacetimeDBClient.cs`](./src/SpacetimeDBClient.cs). This is a general pattern. Similar inheritance patterns are used for tables and indexes: the generated code defines a class that inherits most of its behavior from a class in the SDK.
+
+We require that **a DbConnection is only accessed from a single thread**, which should call the `DbConnection.FrameTick()` method frequently. See [threading model](#threading-model), below.
```
gefjon, line 37: We do? Ew...

DEVELOP.md
```diff
@@ -17,3 +14,85 @@ dotnet pack ../SpacetimeDB/crates/bindings-csharp/BSATN.Runtime && ./tools~/writ
...
+
+Codegen also generates code for each table implementing the `ITable` interface in [`src/Table.cs`](./src/Table.cs). `DbConnection` only sees tables as `ITable`s -- it does not know anything more about the specific implementation of each table. `RemoteTableHandle<...>` in `src/Table.cs` implements the `ITable` interface in combination with the generated code. It also has a callback structure, `OnInternalInsert` and `OnInternalDelete`, used by generated code to maintain indexes.
+
+Roughly speaking, code pertaining to specific tables should live in `Table.cs`, and code pertaining to the connection as a whole should live in `DbConnectionBase<...>`.
+
+### Threading model
+
+The C# SDK, unlike the [Rust SDK](https://github.com/clockworklabs/SpacetimeDB/tree/master/crates/sdk), **assumes a DbConnection is only accessed from a single thread**. This thread is referred to as the "main thread". The "main thread" is:
+- Whichever thread is repeatedly calling `DbConnection.FrameTick()` in a loop.
+It is **only safe to call `FrameTick()` from a single thread**. It is **only safe to access the DbConnection from this thread**.
```
gefjon, line 65: Is this solvable/fixable with a lock?

DEVELOP.md
```diff
@@ -17,3 +14,85 @@ dotnet pack ../SpacetimeDB/crates/bindings-csharp/BSATN.Runtime && ./tools~/writ
...
+`tools~/gen-client-api.bat` script.
+
+Note that messages are actually double-encoded. The `SpacetimeDB.ClientApi` messages store various `byte[]`s that must be decoded *again* to get actual table rows, reducer arguments, etc. This unfortunately involves a lot of copying.
+
+### Overlapping subscriptions
+
+The user may subscribe to the same row in multiple ways, e.g. `SELECT * FROM students WHERE student.age > 5` and `SELECT * FROM students WHERE student.class = 4`. If the user subscribes to both of these queries, the server will send multiple copies of all students in class 4 of age greater than 5.
+
+We could deduplicate queries server-side, but this represents a large amount of work, so for performance reasons we deduplicate them client-side instead. We rely on the [`MultiDictionary`](src/MultiDictionary.cs) class to do this. This class is like a regular dictionary, but it can store multiple "copies" of a (key, value) pair. See the comments on that class for more information, and [`tests~/MultiDictionaryTests.cs`](./tests~/MultiDictionaryTests.cs) for randomized tests of its behavior.
```
gefjon:
```suggestion
We could deduplicate multiply-subscribed rows server-side, but this represents a large amount of work, so for performance reasons we deduplicate them client-side instead. We rely on the [`MultiDictionary`](src/MultiDictionary.cs) class to do this. This class is like a regular dictionary, but it can store multiple "copies" of a (key, value) pair. See the comments on that class for more information, and [`tests~/MultiDictionaryTests.cs`](./tests~/MultiDictionaryTests.cs) for randomized tests of its behavior.
```

We actually do deduplicate queries on the server side, when a client has multiple subscriptions to the same SQL query.

gefjon (approved):
One suggestion and two responses that don't require action. Thanks for writing this, it's gonna be really helpful!

DEVELOP.md
kazimuth: Would require locking on all accesses of `conn.Db`. I don't think there's a nice way to do this without an API break. The type system is just much less expressive RE: sharing.

kazimuth: Even if you, say, require `conn.Db` to be accessed from a callback that has a lock on the database, the user can still just store the Db you passed outside the callback. In the end I think the only way to do it would be to have the user manage the lock.

We could think about doing something MVCC-y, where each `conn.Db` you see is actually immutable, and you're just getting new copies of the Db on changes, but I would be worried about the perf implications there.

kazimuth: That *might* allow more work to be done in the background threads, but it would naively involve a lot more allocations since you'd be creating a new copy of the indices each frame. It might work to just have pools of `Dictionary<...>` lying around that we reuse. Alternatively, we find some sort of haskelly immutable dictionary that can share state between copies.

rekhoff (approved):

bfops:
Migrated to https://github.com/clockworklabs/SpacetimeDB/pull/3140, as we are merging this repo into that one.

Issue #352, opened 2025-07-21
Bug: Token is not used to construct Websocket in WebGL


FME849:
https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/blob/aecb141a21b4d1c75920a19a9182e4d06cc7327c/src/Plugins/WebSocket.jslib#L26
Hey, I found that this `authToken` variable from `DbConnectionBuilder` is not used, therefore it always create new `Identity` when client create new connection to DB

bfops:
Thank you for reporting this! We'll work on getting it fixed.

jdetter:
We're gonna wait for the author to try to move the PR that fixes this issue into the SpacetimeDB public repo.

bfops:
Closing since this is marked done

PR #354, opened 2025-07-23
feat: generate WebSocket token from auth

FME849:
## Description of Changes
This is a fix for community bug #352 
- Using the auth token to get a temporary token.
- Add temporary token as a search param in the Websocket uri.

This idea was inspired by the Typescript SDK

## API

## Requires SpacetimeDB PRs
None

## Testsuite


src/Plugins/WebSocket.jslib
```diff
@@ -18,6 +18,46 @@ mergeInto(LibraryManager.library, {
         manager.callbacks.error = errorCallback;
     },
 
+    WebSocket_Generate_Token: function(baseUriPtr, uriPtr, authTokenPtr, callbackPtr) {
```
jdetter, line 21: Instead of having this be a separate function I would just do this by default in the connection setup logic. If the user passes an authentication token we should create the short-lived auth token automatically and then pass that as a query parameter instead of exposing a separate function.

jdetter (changes requested):
Thank you for the contribution! Small bit of feedback on this but I think this is the correct approach. Also, recently we've made our public SpacetimeDB repository our monorepo which means that the C#/UnitySDK lives here now: https://github.com/clockworklabs/SpacetimeDB/tree/master/sdks/csharp.

Would you mind opening up a PR with these changes against the SpacetimeDB repository instead? Thanks! Feel free to assign me for review when the PR is opened.

bfops:
I'm going to close this since it was migrated to https://github.com/clockworklabs/SpacetimeDB/pull/2988.

